PR #1, opened 2016-11-25
some things you need for react native

hortonelectric:
Readme updates only, for people using react native.

The json-loader bit actually seems relevant to anybody building with node 6 npm 3.x

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,25 +9,6 @@ This will produce a file, `./abc.js`, which can be used in a `<script>` tag.
 All sources are in the [JavaScript Standard Style](http://standardjs.com/).
 
 
-## REACT NATIVE
-change webpack.config.js
-```
-module.exports = {
-  entry: './src/abc.js',
-  module: {
-    loaders: [
-    // note the change from json to json-loader here, it might be something you need to do for anything to actually build
-      { test: /\.json$/, loader: 'json-loader' }
-    ]
-  },
-  output: {
-    filename: 'abc.js',
-    // Export the library as a global var:
-    // libraryTarget: "var",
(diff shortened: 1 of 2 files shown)


swansontec:
Thanks for the PR, but I'd rather have things "just work", rather than specify changes in a readme.

Actually, the Webpack build was really intended for web, and should be irrelevant to React Native. The entire core is in ES5 CommonJS format already, so it should be possible to `require` directly with no transformation. On the other hand, I'm getting ready to switch things to ES6 module syntax, but I'll use rollup.js to handle that, making Webpack irrelevant.

hortonelectric:
Check out the PR again, I confirmed the webpack change so people who want
to compile this will actually be able to do so without any error messages
or tweaking, whilst you still have webpack.

As for the commonjs thing, when it comes down to it, a global doesn't work
with every situation. So I marked it as optional config in the readme, in
our case, we are making it a commonjs load for the module in react native.
Would have saved me an hour if it had already been there, but it's not
surprising it wasn't there, since this is the first react native app.

Copy that on ES6, it seems like a small change.

npm test fails. Seemingly because of something stupid. Can I start creating
issues so all of this stuff I am seeing can get addressed after discussion?

swansontec:
As I said before, Webpack was *only* for browser. For React Native, you should not need to build anything at all. The core is already in CommonJS, so you can just grab it directly. The best thing to do would be to just delete the Webpack setup entirely, since nobody is using it anymore.

I see the issue with the unit tests. Yes, we can start using github issues for these things.

swansontec:
I have deleted Webpack entirely from our core. To use the core from within React Native, just add it to the `package.json` deps like normal, and `require` it like any other package.

PR #2, opened 2016-12-10
non-standard HTTP requests are bad foo. Android won't accept that via…

hortonelectric:
… React Native so we remove the body from the request if it's a GET request


swansontec:
Ugh, this change did not pass the unit tests. There were a number of obvious source-formatting errors that killed our CI server. When doing future changes, please run `npm test` before committing/pushing.

PR #4, opened 2017-09-29
maxSpend (with tests)

safitudo:
Tested with react-gui


paullinator:
@safitudo Please have the routine fixed to ONLY use big number strings. We should never be touching javascript numbers in the Core for crypto amounts.

swansontec:
Please see pull request #5, which is just this branch rebased onto the latest develop.

PR #5, opened 2017-10-03
Get max spendable

swansontec:
This is a replacement for pull request #4 . The other branch had merge conflicts and a spelling error, so I rebased it.


src/currencyWallets/api.js
```diff
@@ -43,7 +43,7 @@ const fakeMetadata = {
   notes: ''
 }
 
-const PRECISION = 1 / 1000000000 // 0.000 000 001
+export const PRECISION = '1'
```
swansontec: We shouldn't need `PRECISION` anymore - the max spend should always be absolutely precise, down to the very last Satoshi or Wei.

src/currencyWallets/currency.test.js
```diff
@@ -7,10 +8,16 @@ import {
 } from '../test/fakeCurrency.js'
 import { fakeExchangePlugin } from '../test/fakeExchange.js'
 import { awaitState } from '../util/redux/reaction.js'
-import { assert } from 'chai'
+import { PRECISION } from './api'
+import chai from 'chai'
```
swansontec: Why not `import { assert } from 'chai'`? Or does the `chai-as-promised` mess this up?

src/currencyWallets/currency.test.js
```diff
@@ -137,6 +144,28 @@ describe('currency wallets', function () {
...
+    const store = makeFakeCurrencyStore()
+    store.dispatch({ type: 'SET_BALANCE', payload: 50 })
+
+    const wallet = await makeFakeCurrencyWallet(store)
+    const maxSpendable = await wallet.getMaxSpendable({
+      currencyCode: 'TEST',
+      spendTargets: [{}]
+    })
+
+    assert.isFulfilled(
```
swansontec: These don't actually pass the flow type checker. I'm not sure how to fix this - `chai-as-promised` seems to not have flow typings. I suppose we could just do `assert.equals(await wallet.getMaxSpendable(...), someAmount)`, bypassing the whole `chai-as-promised` library. If you can find flow typings for it instead, that would be nice too.

src/test/fakeCurrency.js
```diff
@@ -49,6 +50,10 @@ class FakeCurrencyEngine {
       createReaction(state => state.blockHeight, onBlockHeightChanged)
     )
 
+    this.MAX_SPENDABLE_SHARE = 0.9 + 0.1 * Math.random()
```
swansontec: Wouldn't it be better to make the test deterministic, loading different balances into the fake wallet and asserting that the `getMaxSpendable` returns the right value?

You can use `store.dispatch({ type: 'SET_BALANCE', payload: 20 })` to change the balance. There are examples of how to do this all over the `currencyWallet.test.js` file.

src/currencyWallets/currency.test.js
koretskiyav: yes, but on the next line i want to check that promise is rejected and i cant do this without chai plugins

src/test/fakeCurrency.js
paullinator: @koretskiyav What's the status on this PR? I see you have another branch `get_max_pendable` that seems to be newer but there's no PR for it.

swansontec:
Merged into develop.

PR #6, opened 2017-10-06
Pixie wallet

swansontec:


swansontec:
I've tested this locally, and it seems to work at least as well as what we had before. Would @thehobbit85 or @wolverineks be able to review it? I realize this is an absolutely massive change, so I'm OK just merging if nobody has time.

src/modules/exchange/updateExchange.js
```diff
@@ -10,9 +10,7 @@ export function exchangePixie (input: PixieInput<RootProps>) {
 
   return {
     update (props: RootProps): Promise<void> | void {
-      function doFetch (): Promise<void> {
-        const plugins: Array<AbcExchangePlugin> = props.output.exchangePlugins
-
+      function doFetch (plugins: Array<AbcExchangePlugin>): Promise<void> {
```
paullinator: doFetch in the recursion doesn't have parameters

src/modules/context/context-api-pixie.js
```diff
@@ -19,20 +17,19 @@ import {
...
-        ? opts.accountType.replace(/^account.repo:/, '')
-        : ''
+export const contextApiPixie = (input: PixieInput<RootProps>) => (
+  props: RootProps
+) => {
+  input.onOutput(makeContext(input))
+  return stopUpdates
+}
+
+export function makeContext (propsWrapper: PixieInput<RootProps>) {
```
paullinator: Why is this exported? Who imports this makeContext? From this file, it looks like makeContext changed from being a function that accepts an AbcContextOptions to one that takes PixieInput<RootProps>. But no other code changed to complement this switch.

paullinator:
As you said, William, this is too large of a change to grasp by most anyone other than you. I was only able to spot check isolated logic. As we need this change in now, I'll merge this. Let's do PRs from here on out for `airbitz-core-js`, and keep the PRs to smaller changes.

PR #7, opened 2017-10-12
add shapeshift getExchangeSwapRate api call

koretskiyav:


src/modules/currencyWallets/api.js
```diff
@@ -85,6 +89,9 @@ export function makeCurrencyApi (
   const engine = () => getCurrencyWalletEngine(ai.props.state, keyId)
   const plugin = () => getCurrencyWalletPlugin(ai.props.state, keyId)
 
+  const io = getIo(ai.props.state)
```
swansontec: This should just be `ai.props.io`. The whole `getIo` thing is being phased out.

swansontec: Actually, just `const shapeshiftApi = makeShapeshiftApi(ai.props.io)`. Keep it simple. :)

src/modules/exchange/exchangeApi.js
```diff
@@ -43,3 +45,22 @@ function makeExchangeCacheApi (ai: ApiInput) {
...
+}
+
+const API_PREFIX = 'https://shapeshift.io'
+
+export function makeShapeshiftApi (io: FixedIo) {
+  return {
+    async getExchangeSwapRate (fromCurrency: string, toCurrency: string) {
+      const pair = `${fromCurrency}_${toCurrency}`
+      const reply = await io.fetch(`${API_PREFIX}/rate/${pair}`)
+      const json: ExchangeSwapRate = await reply.json()
```
swansontec: When I tested this with a fake server, I got back a 404 error like I expected. However, the 404 included valid JSON as its message body. Therefore, this code returned the error JSON as if it were the exchange data. We eventually need some `if (!reply.ok) { throw new Error(`The server returned HTTP error code ${reply.status}`) }` logic.

swansontec:
I went ahead and pushed this to `develop`. The error-handling issue still needs to be fixed.

PR #8, opened 2017-10-13
Edge Lobby API

swansontec:
This works in the unit-tests. Once we hook up the GUI, we can test it against Augur or some other partner apps. I suppose I could hook up a CLI command for testing as well, if we felt the need.

### DIFF
--- a/src/modules/account/lobbyApi.js
+++ b/src/modules/account/lobbyApi.js
@@ -4,6 +4,7 @@ import { wrapObject } from '../../util/api.js'
 import { base64 } from '../../util/encoding.js'
 import { fetchLobbyRequest, sendLobbyReply } from '../login/lobby.js'
 import type { LobbyRequest } from '../login/lobby.js'
+import { sanitizeLoginStash } from '../login/login.js'
 import type { ApiInput } from '../root.js'
 import { ensureAccountExists, findAppLogin } from './accountState.js'
 
@@ -49,8 +50,7 @@ async function approveLoginRequest (
   const stashTree = await ai.props.loginStore.load(
     accountState.loginTree.username
   )
-  // TODO: Sanitize
-  const loginStash = stashTree
+  const loginStash = sanitizeLoginStash(stashTree, appId)
 
   // Send the reply:
   const replyData = {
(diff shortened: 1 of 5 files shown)


swansontec:
The type changes in airbitz/airbitz-core-types#1 above are needed as well.

swansontec:
I've added a third commit that handles `appId` lookups more gracefully. If the server isn't working, the worst that will happen is that the appId shows up as the app name.

src/modules/account/lobbyApi.js
```diff
@@ -0,0 +1,102 @@
...
+  if (lobbyJson.loginRequest) {
+    const appId = lobbyJson.loginRequest.appId
+    if (typeof appId !== 'string') throw new TypeError('Invalid login request')
+    const { displayName, displayImageUrl } = await fetchAppIdInfo(ai, appId)
+
+    // Make the API:
+    const rawLoginRequest: AbcLoginRequest = {
+      appId,
+      displayName,
+      approve () {
```
paullinator: I know I should have spoken up sooner, but I just realized this is an anti pattern of our API direction. We've generally moved methods out of smaller objects such as this one. AbcContext, AbcAccount, and AbcWallet still have methods but recall how we moved methods out of AbcSpendTarget and smaller temporary objects. Should we do the same here? Ugh, sorry I really wish I said something sooner.

swansontec: The size of the object isn't an issue; it's whether the object belongs to the user or to the core:

* API objects - belong by the core. These can have methods and read-only data.
* User objects - belong to the GUI. These can have read / write data, but no methods.

So, we took methods off of `ABCReceiveAddress`, for example, because the user was supposed to edit the metadata to do an update. Thus, the object belongs to the user and cannot have methods.

This AbcLoginRequest is solidly a core object. The core creates it, and none of the properties are writable (the wrapApi call locks them all down). Thus, I see no problem with giving it a method.

paullinator:
Won't the App have to modify the request in the future when it decides to share preexisting wallets as part of the login? In that sense the object would be modified by the user but then still have methods on it.

swansontec:
No. The `approve` method would take some type of parameter to indicate the wallets. So, it would look like this:

```js
loginReqest.approve(myWalletId)
```

As opposed to:

```js
loginRequest.replyWallet = myWallet
loginRequest.approve()
```

The second option involves writing to the login request object, and would necessitate moving `approve` somewhere else. Fortunately, we don't have to do that.

PR #9, opened 2017-10-16
Exchange shapeshift api

koretskiyav:


env.example.js
```diff
@@ -0,0 +1,3 @@
+export default {
+  SHAPESIFT_PUBLIC_API_KEY: 'SHAPESIFT_PUBLIC_API_KEY'
```
swansontec: Spelling mistake.

src/modules/currencyWallets/api.js
```diff
@@ -17,8 +17,8 @@ import {
   setCurrencyWalletTxMetadata,
   setupNewTxMetadata
 } from '../actions.js'
-import { makeShapeshiftApi } from '../exchange/exchangeApi.js'
-import type { ExchangeSwapRate } from '../exchange/exchangeApi.js'
+import makeShapeshiftApi from '../exchange/api.js'
```
swansontec: It's really hard to debug filenames like this, since you're never sure which `api.js` you want when there are several with the same name. A better name would be `shapeshift.js`.

src/modules/exchange/api.js
```diff
@@ -0,0 +1,48 @@
...
+import ENV from '../../../env'
+
+export type ExchangeSwapRate = {
+  pair: string,
+  rate: string,
+}
+
+const API_PREFIX = 'https://shapeshift.io'
+
+export default function makeShapeshiftApi (io: FixedIo) {
```
swansontec: We tend to avoid default exports, but we can let this slide for now.

src/modules/exchange/api.js
```diff
@@ -0,0 +1,48 @@
+// @flow
+import type { FixedIo } from '../../io/fixIo.js'
+
+import ENV from '../../../env'
```
swansontec: The core is supposed to be configuration-free. We accept all our API keys and stuff through the `makeContext` call. We can fix this later, since we're in a hurry, but it is a problem.

swansontec: Actually, by the time we get this fixed to work on our build server, it would have been faster to just fix it in the first place. We need to get this taken care of.

swansontec:
I merged this to develop. @koretskiyav, if you need to make any changes in the future, please start a new branch off of develop.

PR #12, opened 2017-10-19
fix Max Send for Wings/REP

koretskiyav:


src/modules/currencyWallets/api.js
```diff
@@ -382,6 +382,7 @@ export function makeCurrencyApi (
```
paullinator: line 380.. should this be `lt(sub(max, min)`.... Not `lte`. If min and max are still different by one then you need to try both options to see which fails otherwise you'll have an off by one error.

src/modules/currencyWallets/api.js
```diff
@@ -390,7 +391,7 @@ export function makeCurrencyApi (
         }
       }
 
-      return getMax('0', balance)
+      return getMax('0', add(balance, '1'))
```
paullinator: Not sure why 1 was added here. Can you explain? Is this to compensate for the `lte` above?

koretskiyav: Yes, but only for [balance - 1, balance] case. My bad, fixed in next commit.

PR #16, opened 2017-10-23
Do not process source files through babel es2015 for react-native

paullinator:
On Android we hit a problem whereby es2015 babel plugin will create nested functions inside IF statements which causes an error in RN since "use strict" is on by default. This change creates a separate build for react native that is only processed for flow and spread operators. This fixes Android and tested to still work on iOS. Performance on ios appears unchanged from prior to this PR

### DIFF
--- a/package.json
+++ b/package.json
@@ -26,12 +26,14 @@
   ],
   "main": "lib/indexABC.js",
   "module": "lib/indexABC.es.js",
+  "react-native": "lib/indexABC.rn.js",
   "repository": {
     "type": "git",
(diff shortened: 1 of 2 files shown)


swansontec:
As mentioned on slack, React Native has a simple config difference in their `.babelrc` compared to ours, which explains why our code fails in their environment. They set the "loose" flag to true on the for-of syntax transformer, which bypasses all the weird iterator stuff. We should revert this change and try that instead some day, but for now it's just nice to have things working.

PR #17, opened 2017-10-23
Wallet recursion

swansontec:
With this, we can finally see logins from partner apps!


swansontec:
Please don't merge this. I have found some issues testing with the real app.

swansontec:
Found the problem; everything seems to work well now.

swansontec:
@thehobbit85, can you take a look at this one?

PR #19, opened 2017-10-24
Include the storage wallet instantiation function into distribution files

jtrancas:
According to the documentation at https://developer.airbitz.co/javascript/#makestoragewallet the makeStorageWallet() function should be available in order to instantiate an ABCStorageWallet, however, the rollup is not including this method on the distribution code during the build process. The specified method was imported to the entry point file so rollup can include it during the build.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -1,4 +1,5 @@
 import { makeCurrencyWallet } from './currencyWallets/api.js'
+import { makeStorageWallet } from "./storage/storageApi";
 import { makeContext } from './io/context.js'
 
 // Sub-module exports:
@@ -36,3 +37,8 @@ export function makeABCContext (apiKey, appId, opts) {
  * Creates a new wallet object based on a set of keys.
  */
 export { makeCurrencyWallet }
+
+/**
+ * Creates a new wallet object based on a set of keys.
+ */
+export { makeStorageWallet}


swansontec:
Thanks for the pull request! Yes, this is definitely an oversight.

Currently, all our development is happening in the "develop" branch. We then merge master to develop when we do a release. Unfortunately, on `develop`, we have made some changes to the way wallet management works, and we haven't updated the docs to match. At least for currency wallets, the account now manages the wallet lifetimes based on which currency plugins are enabled. We are considering doing the same for storage wallets, where you provide a list of wallet types you want to use for storage, and we automatically boot them up and return them in an array.

So, I'm not really sure how to move forward with this. We could merge to master, but then we wouldn't be doing an NPM release any time soon. Can you tell me a little more about what you're working on, so we can figure out how best to help you proceed?

jtrancas:
@swansontec Thanks for the reply, I actually noticed that you were actively working on the develop branch, but I wasn't aware of your roadmap for the next releases so I stil did the pull request because right now I really need to make use of storage wallets.

Currently I'm working on a web app that depends on airbitz-core-js login/register/recovery api's as well as on the storage wallets to store some (currently unsupported) blockchain keys, we're also using the account wallet to store some account related data. Are these APIs going to suffer fundamental changes? And what are your plans for the next release with the updated documention?

swansontec:
The API isn't changing fundamentally. Once you get a storage wallet, it will have the same methods and callbacks. The only thing we might change is the way you *get* a storage wallet, hopefully making it simpler. At the very least, we would move the `makeStorageWallet` API onto the account or context object instead of leaving it loose.

Even if Airbitz doesn't understand your blockchain keys, we still *strongly* suggest keeping them in the account wallet key list. Just let us know what the wallet type is, so we can add it to the docs. The account key storage mechanism is append-only, while the git-based storage supports deletes and modifications. Since nobody wants to lose keys, the account key list is much safer.

Do you need us to do an NPM publish on the master branch? Otherwise you can just do "npm install git://github.com/jtrancas/airbitz-core-types.git", which I assume you are already doing.

jtrancas:
The blockchain I'm dealing with is BitShares, do you have plans to include support for it?
When you say I should use the account key storage mechanism, you mean I should use a CurrencyWallet, right? Which means that support for bitshares must be provided. Or is there a different way of adding those keys?

Right now I can use npm install <my github repo url> but it's far from perfect because when I install it it doesn't seem to run the "prepare" script, so the dist files are not generated, according to their docs the npm install should run the "prepare" script since npm 5.x.x, however it doesn't seem to be working.

swansontec:
The `EdgeAccount.dataStore` now provides an officially-supported method of storing arbitrary data in an Edge account. We suggest using this going forward, for anybody interested.

PR #21, opened 2017-10-27
prevents error in cases where out.metadata is undefined

thehobbit85:


swansontec:
This seems to be an on-disk corruption issue (we have a file, and that file has `metadata: null` or something in the JSON). I'll pull the if statement up a bit higher in the function, since it could mess with other stuff.

swansontec:
Er, nope. It is a token vs. primary-currency issue. I'll still move the check higher in the file, and add a comment too.

swansontec:
I have made a better version in #22. This fix may stop the core from crashing, but I expect the GUI would probably crash in this case.

PR #22, opened 2017-10-27
Never return transactions without metadata

swansontec:
If there is no metadata on disk, just return some fallback data.


swansontec:
This is a more thorough version of #21

PR #24, opened 2017-11-01
Legacy metadata

swansontec:
With this in place, logging in with legacy accounts will no longer be dangerous. The user's metadata will not be destroyed or lost.

### DIFF
--- a/src/modules/currency/wallet/currency-wallet-files.js
+++ b/src/modules/currency/wallet/currency-wallet-files.js
@@ -10,9 +10,28 @@ import {
 import { getCurrencyMultiplier } from '../currency-selectors.js'
 import type { CurrencyWalletInput } from './currency-wallet-pixie.js'
 
-function getTxFile (state, keyId, timestamp, txid) {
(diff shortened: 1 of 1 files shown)


src/modules/currency/wallet/currency-wallet-files.js
```diff
@@ -66,74 +143,112 @@ export function setCurrencyWalletFiat (
...
+  // Save the results to redux:
+  return allFiles.then(allFiles => {
+    const [newFiles, oldFiles] = allFiles
+
+    const out = {}
+    for (const json of newFiles) {
+      if (json == null || !json.txid) continue
+      out[json.txid] = json
+    }
+    for (const json of oldFiles) {
```
paullinator: This seems like you will overwrite data from newFiles with data from oldFiles. Since this runs in Edge, shouldn't you prioritize newFiles over oldFiles?

swansontec:
Good catch; I flipped it around for testing, but forgot to flip it back. Fix pushed.

PR #31, opened 2017-12-26
Add getExchangeSwapInfo

wolverineks:


src/modules/context/context-api-pixie.js
```diff
@@ -194,6 +194,19 @@ function makeContextApi (ai: ApiInput) {
       toCurrencyCode: string
     ): Promise<number> {
       return shapeshiftApi.getExchangeSwapRate(fromCurrencyCode, toCurrencyCode)
+    },
+
+    getExchangeSwapInfo (
+      fromCurrencyCode: string,
+      toCurrencyCode: string
+    ): Promise<{
```
swansontec: Can you turn this into a proper data type, and document what the fields mean? I would like to see it be a little more consistent too, since we are not just tied to Shapeshift (we want other partners in the future).

```js
export type ExchangeSwapInfo = {
  rate: number,
  min: string, // Native amount, as string
  max: string // Native amount, as string
}
```

I left off the minerFee and pair. We already know what the pair is, and I have no idea what the `minerFee` is for.

src/modules/exchange/shapeshift.js
```diff
@@ -51,6 +51,21 @@ export function makeShapeshiftApi (ai: ApiInput) {
...
+    ): Promise<{
+      pair: string,
+      rate: number,
+      limit: number,
+      min: number,
+      minerFee: number
+    }> {
+      const pair = `${fromCurrency}_${toCurrency}`
+      const json = await api.get(`/marketinfo/${pair}`)
+      return json
```
swansontec: Assuming the format uses nativeAmounts, we need to process the JSON into strings.

PR #32, opened 2017-12-28
2fa

swansontec:
Add 2fa management endpoints


ahujsak:
Ignoring travis at our own peril based on Williams input,... But go ahead and merge.

PR #49, opened 2018-02-16
Add getAvailableExchangeTokens to context

ahujsak:
• gets tokens supported by shapeshift
• loops through and returns only available ones


swansontec:
Included in core version 0.5.6.

PR #50, opened 2018-02-24
Bump sync server timeout from 10s to 3 minutes

paullinator:
This allows larger sync repos to get synced before the timeout expires

### DIFF
--- a/src/modules/storage/servers.js
+++ b/src/modules/storage/servers.js
@@ -1,5 +1,4 @@
 import { NetworkError } from '../../error.js'
-import { timeout } from '../../util/promise.js'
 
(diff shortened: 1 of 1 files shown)


swansontec:
We should just delete the `timeout` call entirely, rather than change the number. This way, the OS network stack can manage the connection with its own proper timeouts.

paullinator:
Is there a way for us to use the timeout to at least generate the notification that the sync is taking a long time. Basically every 10s generate an onError with “Still synchronizing wallet...”

swansontec:
That's a whole different ball of wax, but it is possible. It would be some sort of setTimeout at the start of the sync, and a clearTimeout at the end of the sync. If the timer goes off, we call onError.

PR #51, opened 2018-02-27
Undelete split

swansontec:
This is completely untested code! I just want to push this up to show you what I was thinking.

### DIFF
--- a/src/modules/account/accountState.js
+++ b/src/modules/account/accountState.js
@@ -455,7 +455,10 @@ class AccountState {
 
     const newWalletInfo = splitWalletInfo(walletInfo, newWalletType)
     if (this.allKeys.find(walletInfo => walletInfo.id === newWalletInfo.id)) {
-      throw new Error('This wallet has already been split')
+      const walletInfos = {}
+      walletInfos[walletInfo.id] = { deleted: false }
+      await this.changeKeyStates(walletInfos)
+      return walletInfo.id
     }
 
     // Add the keys to the login:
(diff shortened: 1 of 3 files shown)


src/modules/account/accountState.js
```diff
@@ -483,6 +483,20 @@ class AccountState {
...
   }
 
+  listSplittableWalletTypes (walletId) {
+    const walletInfo = this.allKeys.find(
+      walletInfo => walletInfo.id === walletId
+    )
+    if (!walletInfo) throw new Error(`Invalid wallet id ${walletInfo.id}`)
+
+    // This is the bare minimum for now:
+    if (walletInfo.type === 'wallet:bitcoin') {
```
swansontec: Probably need something like ` && walletInfo.keys.type !== 'bip49'`

thehobbit85: Do you have access to the pluginCurrency info here?

swansontec: I believe we can get access to the currencyInfo from there, yes. It should be `this.ai.props.state.currency.infos`. I think there is also a selector, `getCurrencyInfo`, so you could do `getCurrencyInfo(this.ai.props.state, walletType)`

thehobbit85: If you access to the currencyInfos then we can do something like this:

```javascript
return pluginCurrencyInfoToSplitFrom.forks.filter(walletType => {
   pluginCurrencyInfoToSplitTo.formats.find(walletInfo.keys.format)
})
```

And I'll add 'formats' and 'forks' into the currencyInfos

PR #53, opened 2018-02-28
Prevent GUI crashes due to corrupt metadata.amountFiat

paullinator:
Check if fiat amount exceeds standard JS precision. If so, zero the amount before returning to GUI.

### DIFF
--- a/src/modules/currency/wallet/currency-wallet-api.js
+++ b/src/modules/currency/wallet/currency-wallet-api.js
@@ -416,6 +416,10 @@ export function combineTxWithFile (
     merged.metadata.exchangeAmount[walletFiat]
   ) {
     out.metadata.amountFiat = merged.metadata.exchangeAmount[walletFiat]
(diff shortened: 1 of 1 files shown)


src/modules/currency/wallet/currency-wallet-api.js
```diff
@@ -416,6 +416,10 @@ export function combineTxWithFile (
     merged.metadata.exchangeAmount[walletFiat]
   ) {
     out.metadata.amountFiat = merged.metadata.exchangeAmount[walletFiat]
+    if (out.metadata.amountFiat.toString().includes('e')) {
+      // Corrupt amountFiat that exceeds a number that JS can cleanly represent without exponents. Set to 0
+      out.metadata.amountFiat = 0
```
thehobbit85: Can you try and round the number up in any way to still get a meaningful result?

paullinator:
Eliran claims to approve (via phone)

PR #55, opened 2018-03-01
List splittable types

thehobbit85:


src/edge-core-index.js
```diff
@@ -397,6 +397,7 @@ export type EdgeAccount = {
   getFirstWalletInfo(type: string): ?EdgeWalletInfo,
   getWalletInfo(id: string): EdgeWalletInfo,
   listWalletIds(): Array<string>,
+  listSplittableWalletTypes(walletId: string): Array<string>,
```
swansontec: We still need to add this function to the `EdgeCurrencyPlugin` type.

src/modules/account/accountState.js
```diff
@@ -483,6 +483,23 @@ class AccountState {
...
 
+  listSplittableWalletTypes (walletId) {
+    const walletInfo = this.allKeys.find(
+      walletInfo => walletInfo.id === walletId
+    )
+    if (!walletInfo) throw new Error(`Invalid wallet id ${walletInfo.id}`)
+    const plugin = getCurrencyPlugin(
+      this.ai.props.output.currency.plugins,
+      walletInfo.type
+    )
```
swansontec: There is no error handling for if `plugin` is `undefined`.

src/edge-core-index.js
thehobbit85: added

src/modules/account/accountState.js
thehobbit85: added

PR #56, opened 2018-03-05
Pass io object into syncStorageWallet

paullinator:
Prevent sync failure due to lack of io object


swansontec:
Waiting for travis to finish, but looks good.

PR #58, opened 2018-03-07
Get-Transaction optimize performance

thehobbit85:
Use a much more "lazy" approach for loading the transaction data from disk.
We are currently loading all of transaction data from disk to memory when on login.
The problem with that, is that loading the TX data from disk requires decrypting which takes tons of CPU cycles and time and also we really don't need the TX until we move into the transaction list scene.

The main changes are:
1. On startup, load only the TX file names without decrypting them.
2. Only load and decrypt the TX data if the user ask the TX using the "getTransaction" method.
3. When getting a TX from the plugin, call "onTransactionChange" only if we never created a file which removes tons of pointless calls.

### DIFF
--- a/src/modules/actions.js
+++ b/src/modules/actions.js
@@ -95,7 +95,7 @@ export interface CurrencyWalletFileChanged {
     json: any,
     txid: string,
     walletId: string,
-    txidHash: string
+    txFileName: any
   };
 }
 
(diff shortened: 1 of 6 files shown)


src/modules/currency/wallet/currency-wallet-reducer.js
```diff
@@ -71,6 +81,47 @@ const currencyWalletReducer = buildReducer({
     return action.type === 'CURRENCY_WALLET_FILES_LOADED' ? true : state
   },
 
+  sortedTransactions (state = [], action: RootAction, next: CurrencyWalletNext) {
+    switch (action.type) {
+      case 'CURRENCY_ENGINE_CHANGED_TXS': {
```
kylanhurt: At some point you may want to change these strings to constants.

src/modules/actions.js
```diff
@@ -33,7 +33,8 @@ export interface CurrencyEngineChangedTxs {
   type: 'CURRENCY_ENGINE_CHANGED_TXS';
   payload: {
     txs: Array<any>,
-    walletId: string
+    walletId: string,
+    txidHashes: any
```
kylanhurt: I have a feeling this will need to be changed to something like `Array<string>`?

PR #65, opened 2018-04-10
Allen/shape shift preceise

ahujsak:


src/edge-core-index.js
```diff
@@ -643,6 +644,34 @@ export type EdgeReceiveAddress = EdgeFreshAddress & {
   nativeAmount: string
 }
 
+export type EdgeCoinExchangeQuote = {
+  depositAmount: string,
+  depositAmountNative: string,
+  withdrawalAmount: string,
+  withdrawalAmountNative: string,
+  expiration: number,
```
swansontec: // ms

src/edge-core-index.js
```diff
@@ -643,6 +644,34 @@ export type EdgeReceiveAddress = EdgeFreshAddress & {
   nativeAmount: string
 }
 
+export type EdgeCoinExchangeQuote = {
+  depositAmount: string,
```
swansontec: Needed?

src/edge-core-index.js
```diff
@@ -643,6 +644,34 @@ export type EdgeReceiveAddress = EdgeFreshAddress & {
...
+  withdrawalAmount: string,
+  withdrawalAmountNative: string,
+  expiration: number,
+  quotedRate: string,
+  maxLimit: number,
+  orderId: string,
+  edgeTransacton: EdgeTransaction
+}
+
+export type SSExchangeQuote = {
```
swansontec: Move into shapeshift.js, not needed externally

src/error.js
```diff
@@ -112,3 +113,14 @@ export function UsernameError (message = 'Invalid username') {
...
 }
+
+/**
+ * Attempting to shape shift between two wallets of same currency.
+ */
+export function SameCurrencyError (
+  message = 'Wallets can not be the same currency'
+) {
+  const e = new Error(message)
+  e.name = e.type = errorNames.SameCurrencyError
```
swansontec: Get rid of e.type, this is an old thing we need to stop doing.

PR #67, opened 2018-04-13
Sync the account before & after edge logins

swansontec:


superhero2017:
Is it possible to get logged accounts as a synced response?

swansontec:
> Is it possible to get logged accounts as a synced response?

I don't know what you are asking. If you want to know which usernames have logged in on this device, look at `EdgeContext.localUsers` (https://github.com/EdgeApp/edge-core-js/blob/4bf7864c114f1f020bdf8ea2f2b770697aa1ad41/src/types/types.js#L951).

PR #69, opened 2018-04-18
Throttle calls to onTransactionsChanged and onNewTransactions to max every 5 seconds

paullinator:
This prevents the GUI from getting hammered with updates due to initial login which then cause calls to getTransactions over and over which are right now expensive due to the need to decrypt data off disk.

Change unit tests to wait after the first update as onTransactionsChanged and onNewTransactions may not be called synchronously with the update but rather after a setTimeout launched in the future.

Add unit tests which make sure back to back update calls get batched into one callback.

### DIFF
--- a/src/modules/currency/wallet/currency-wallet-callbacks.js
+++ b/src/modules/currency/wallet/currency-wallet-callbacks.js
@@ -40,6 +40,64 @@ export function forEachListener (
   }
 }
 
+type ThrottleDetails = {
+  [walletId: string]: {
+    delayCallback: boolean,
+    lastCallbackTime: number,
+    txArray: Array<EdgeTransaction>
+  }
+}
+
+const throttleDetailsNewTx: ThrottleDetails = {}
+const throttleDetailsChangedTx: ThrottleDetails = {}
+
+let throttleRateLimitMs = 5000
+
+function throttledTxCallback (
+  throttleDetails: ThrottleDetails,
+  callback: Function,
+  walletId: string,
+  txArray: Array<EdgeTransaction>
+) {
+  // If this is a unit test, lower throttling to 250ms
+  if (txArray[0].txid.length < 3) {
+    throttleRateLimitMs = 250
+  }
+  if (!throttleDetails[walletId]) {
+    throttleDetails[walletId] = {
+      delayCallback: false,
+      lastCallbackTime: 0,
+      txArray: []
+    }
+  }
+  if (!throttleDetails[walletId].delayCallback) {
+    const now = Date.now()
+    if (
+      now - throttleDetails[walletId].lastCallbackTime >
+      throttleRateLimitMs
+    ) {
+      callback(walletId, txArray)
+      throttleDetails[walletId].txArray = []
+      throttleDetails[walletId].lastCallbackTime = Date.now()
+    } else {
+      console.log(
+        'throttledTxCallback setTimeout delay callback id:' + walletId
+      )
+      throttleDetails[walletId].delayCallback = true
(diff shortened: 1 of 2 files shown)


src/modules/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -40,6 +40,64 @@ export function forEachListener (
...
+      throttleDetails[walletId].txArray = txArray
+      setTimeout(() => {
+        callback(walletId, throttleDetails[walletId].txArray)
+        throttleDetails[walletId].delayCallback = false
+        throttleDetails[walletId].txArray = []
+        throttleDetails[walletId].lastCallbackTime = Date.now()
+      }, throttleRateLimitMs)
+    }
+  } else {
+    Array.prototype.push.apply(throttleDetails[walletId].txArray, txArray)
```
swansontec: Better syntax is `throttleDetails[walletId].txArray.push(...txArray)` - it transpiles to exactly what you typed, but is a lot more understandable.

src/modules/currency/wallet/currency-wallet.test.js
```diff
@@ -106,20 +106,39 @@ describe('currency wallets', function () {
...
+            txState = [{ txid: 'f', nativeAmount: '200' }]
+            store.dispatch({ type: 'SET_TXS', payload: txState })
+            txState = [{ txid: 'g', nativeAmount: '200' }]
+            store.dispatch({ type: 'SET_TXS', payload: txState })
+            setTimeout(() => {
+              log.assert(['new e', 'new f', 'new g'])
+            }, 251)
+          }, 251)
+        }, 251)
+      }, 251)
```
swansontec: 251 seems really close to 250. Knowing how Travis / Jenkins are, this will probably fail randomly from time to time.

src/modules/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -40,6 +40,64 @@ export function forEachListener (
...
+      now - throttleDetails[walletId].lastCallbackTime >
+      throttleRateLimitMs
+    ) {
+      callback(walletId, txArray)
+      throttleDetails[walletId].txArray = []
+      throttleDetails[walletId].lastCallbackTime = Date.now()
+    } else {
+      console.log(
+        'throttledTxCallback setTimeout delay callback id:' + walletId
+      )
```
swansontec: I would like to see `io.console.log` instead of `console.log` Since the core is intended to be used in other projects, we have to be more well-behaved than the GUI, which owns the world. I know the GUI just uses `console.log` everywhere, but the core is still 100% using the io object.

src/modules/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -40,6 +40,64 @@ export function forEachListener (
...
+type ThrottleDetails = {
+  [walletId: string]: {
+    delayCallback: boolean,
+    lastCallbackTime: number,
+    txArray: Array<EdgeTransaction>
+  }
+}
+
+const throttleDetailsNewTx: ThrottleDetails = {}
+const throttleDetailsChangedTx: ThrottleDetails = {}
```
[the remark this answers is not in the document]
swansontec: Actually, the whole callback thing runs inside the currency engine pixie. Therefore, if you just put the variables inside the body of the `makeCurrencyWalletCallbacks`, they will stick around for the lifetime of the currency engine, and will be scoped to that particular wallet! So there is no need to make a table of objects per wallet.

swansontec:
I created a second pull request with a corrected version of the logic.

PR #71, opened 2018-04-19
One-liner fix for getTransactions when query doesn't start at zero

paullinator:
Add numEntries to numIndex to get full range requested


thehobbit85:
There is nothing to fix here.
It was meant to work this way.

swansontec:
Array.slice takes a start and end index. If `numEntries` is actually an index, you'd better fix your variable names. Otherwise, you have a bug. I'm going to merge this.

PR #72, opened 2018-04-20
Add API for sweepPrivateKeys and fix flow type

thehobbit85:


src/edge-core-index.js
```diff
@@ -689,6 +689,7 @@ export type EdgeCurrencyEngine = {
   addGapLimitAddresses(addresses: Array<string>, options: any): void,
   isAddressUsed(address: string, options: any): boolean,
   makeSpend(abcSpendInfo: EdgeSpendInfo): Promise<EdgeTransaction>,
+  sweepPrivateKeys(abcSpendInfo: EdgeSpendInfo): Promise<EdgeTransaction>,
```
swansontec: How do you pass in the private keys to sweep?

src/modules/currency/wallet/currency-wallet-api.js
```diff
@@ -214,6 +214,10 @@ export function makeCurrencyWalletApi (
       return engine.makeSpend(spendInfo)
     },
 
+    async sweepPrivateKeys (spendInfo: EdgeSpendInfo): Promise<EdgeTransaction> {
```
swansontec: This needs to be added to `EdgeCurrencyWallet`, then, not just the engine.

swansontec: The `EdgeCurrencyWallet` already has this method defined in its API: `sweepPrivateKey(keyUri: string): Promise<void>`. At the very least, we should delete the old function. However, I think the API you have is just... wrong. The API we already have makes more sense.

src/edge-core-index.js
```diff
@@ -514,7 +514,10 @@ export type EdgeCurrencyWallet = {
   signTx(tx: EdgeTransaction): Promise<EdgeTransaction>,
   broadcastTx(tx: EdgeTransaction): Promise<EdgeTransaction>,
   saveTx(tx: EdgeTransaction): Promise<void>,
-  sweepPrivateKey(keyUri: string): Promise<void>,
+  sweepPrivateKeys(
+    edgeSpendInfo: EdgeSpendInfo,
+    options?: any
```
swansontec: This needs to be typed. I'm going to approve this, even though it's rubbish, because we need to get this feature out the door.

PR #73, opened 2018-04-25
Add export of QBO and CSV files

paullinator:
This uses pre-existing getTransactions to query transaction range and then generate QBO or CSV files with those transactions. This is a breaking change to the Core API as getTransactions was using incorrect options parameters.

numEntries and numIndex should have been startEntries and startIndex per `edge-docs` repo.

Bump version accordingly


swansontec:
The unit tests are not passing for this feature. Also, although you added 2 new methods to the EdgeCurrencyWallet API, they aren't in the types. I'll fix this up, as well as move the implementations & unit tests to their own files. This is a lot of code to be adding to some already-long files.

swansontec:
All the dates are off by 9 hours. I'm guessing this is because you wrote this in a different time zone from the San Diego office.

swansontec:
All the dates are off by 9 hours. I'm guessing this is because you wrote this in a different time zone from the San Diego office.

PR #75, opened 2018-05-01
Add WebSocket to the io object

swansontec:


src/io/react-native/react-native-io.js
```diff
@@ -73,7 +73,8 @@ export function makeReactNativeIo (): Promise<EdgeRawIo> {
       TLSSocket,
       pbkdf2,
       scrypt,
-      secp256k1
+      secp256k1,
+      WebSocket: window.WebSocket
```
paullinator: Little confused. How does `window.WebSocket` suddenly exist on RN?  Doesn't a polyfill need to be imported?

swansontec: It is available: https://facebook.github.io/react-native/docs/network.html#websocket-support

I tested this by doing a `console.log` of the `io` object under React Native, and the `WebSocket` did indeed look correct.

paullinator: wow. didn't realize it was baked in. So used to having to include external libs. Approved

PR #76, opened 2018-05-01
Add throttling to onAddressesChecked and onBalanceChanged callbacks

paullinator:
Change unit test to use async/await

### DIFF
--- a/src/modules/currency/wallet/currency-wallet-callbacks.js
+++ b/src/modules/currency/wallet/currency-wallet-callbacks.js
@@ -47,20 +47,19 @@ let throttleRateLimitMs = 5000
  * Returns a function that can be called at high frequency, and batches its
  * inputs to only call the real callback every 5 seconds.
  */
-function makeThrottledCallback (
+function makeThrottledCallback<Arg> (
   input: CurrencyWalletInput,
-  callback: (arg: any) => mixed
-) {
+  callback: (arg: Arg) => mixed
+): (callbackArg: Arg) => mixed {
   const walletId = input.props.id
   const { console } = input.props.io
 
   let delayCallback = false
   let lastCallbackTime = 0
-  let finalCallbackArg: any
+  let finalCallbackArg: Arg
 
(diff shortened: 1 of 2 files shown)


src/modules/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -40,6 +40,56 @@ export function forEachListener (
...
+let throttleRateLimitMs = 5000
+
+/**
+ * Wraps a single value accepting callback with throttling logic.
+ * Returns a function that can be called at high frequency, and batches its
+ * inputs to only call the real callback every 5 seconds.
+ */
+function makeThrottledCallback (
+  input: CurrencyWalletInput,
+  callback: (arg: any) => mixed
```
swansontec: I'd make this a template-style type, so we can avoid `any`.

src/modules/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -40,6 +40,56 @@ export function forEachListener (
...
+      finalCallbackArg = callbackArg
+    } else {
+      const now = Date.now()
+      if (now - lastCallbackTime > throttleRateLimitMs) {
+        lastCallbackTime = now
+        callback(callbackArg)
+      } else {
+        console.info(
+          console.info(`makeThrottledCallback delay, walletId: ${walletId}`)
+        )
```
swansontec: The `console.info` call is doubled up.

src/modules/currency/wallet/currency-wallet.test.js
```diff
@@ -14,6 +14,10 @@ import { fakeExchangePlugin } from '../../../fake-plugins/fakeExchange.js'
 import { makeAssertLog } from '../../../util/assertLog.js'
 import { awaitState } from '../../../util/redux/reaction.js'
 
+function snooze (ms: number) {
```
swansontec: Thanks for fixing the tests.

PR #79, opened 2018-05-11
Store txids for future performance optimizations

wolverineks:


src/modules/actions.js
```diff
@@ -38,6 +38,17 @@ export interface CurrencyEngineChangedTxs {
   };
 }
 
+/**
+ * Called when a currency engine fires the onTxidsChanged callback.
+ */
+export interface CurrencyEngineChangedTxids {
+  type: 'CURRENCY_ENGINE_CHANGED_TX_IDS';
```
swansontec: I'd like to see `TXIDS` as one word. Most Bitcoin folks treat `txid` as a single word.

src/edge-core-index.js
```diff
@@ -686,6 +688,7 @@ export type EdgeCurrencyEngine = {
   getBalance(options: any): string,
   getNumTransactions(options: any): number,
   getTransactions(options: any): Promise<Array<EdgeTransaction>>,
+  getTxids(): Array<string>,
```
paullinator: Can you make this optional as not all plugins have this call? Or does the remaining code absolutely need it? Ideally this is an optimization.

src/edge-core-index.js
```diff
@@ -323,6 +323,7 @@ export type EdgeAccountCallbacks = {
     walletId: string,
     abcTransactions: Array<EdgeTransaction>
   ) => mixed,
+  +onTxidsChanged?: (walletId: string, txids: Array<string>) => mixed,
```
paullinator: I don't believe the GUI should need this callback since it already gets the onTransactionsChanged.

src/fake-plugins/fakeCurrency.js
```diff
@@ -99,9 +100,14 @@ class FakeCurrencyEngine {
           onTransactionsChanged(changed)
 
           // Save the new list of transactions:
+          const oldTxids = Object.keys(oldTxs)
           for (const tx of txs) {
             oldTxs[tx.txid] = tx
           }
+          const newTxids = Object.keys(oldTxs)
```
paullinator: Should not be needed by GUI

PR #81, opened 2018-05-18
William/flow account

swansontec:
This is pretty chunky, but it improves typing in a lot of places. It's better to read the pull-request commit-by-commit.

### DIFF
--- a/src/modules/account/account-api.js
+++ b/src/modules/account/account-api.js
@@ -7,6 +7,7 @@ import type {
   EdgeCurrencyWallet,
   EdgeLobby,
   EdgeWalletInfo,
+  EdgeWalletInfoFull,
   EdgeWalletStates
 } from '../../edge-core-index.js'
 import { copyProperties, wrapObject } from '../../util/api.js'
@@ -30,7 +31,7 @@ export function makeAccount (
   appId: string,
   loginTree: LoginTree,
   loginType: string = '',
-  callbacks: EdgeAccountCallbacks | {} = {}
+  callbacks: EdgeAccountCallbacks = {}
 ) {
   return makeAccountState(ai, appId, loginTree, callbacks).then(state =>
     wrapObject('Account', makeAccountApi(state, loginType, callbacks))
@@ -43,7 +44,7 @@ export function makeAccount (
 function makeAccountApi (
   state: AccountState,
   loginType: string,
-  callbacks: EdgeAccountCallbacks | {}
+  callbacks: EdgeAccountCallbacks
 ): EdgeAccount {
   const ai: ApiInput = state.ai
   const { activeLoginId, accountWalletInfo } = state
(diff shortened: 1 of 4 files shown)


swansontec:
@wolverineks Can you take a look at this?

src/modules/login/login.js
```diff
@@ -270,7 +281,11 @@ export function sanitizeLoginStash (stashTree: LoginStash, appId: string) {
     stash => stash.appId === appId,
     stash => stash,
     (stash, children) => {
-      const login = filterObject(stash, ['username', 'appId', 'loginId'])
+      const login: LoginStash = filterObject(stash, [
+        'username',
+        'appId',
+        'loginId'
+      ])
       login.children = children
```
paullinator: Does this need an elvis statement like above?

src/modules/login/login.js
```diff
@@ -253,8 +260,12 @@ export function makeLoginTree (
...
     stash => makeLoginTreeInner(stash, loginKey),
     (stash, children) => {
-      const login = filterObject(stash, ['username', 'appId', 'loginId'])
-      login.children = children
+      const login: LoginTree = filterObject(stash, [
+        'username',
+        'appId',
+        'loginId'
+      ])
+      login.children = children != null ? children : []
```
paullinator: Was this a bug being fixed?

src/modules/account/accountApi.js
```diff
@@ -87,7 +88,7 @@ function makeAccountApi (
 
     // Change or create credentials:
     changePassword (password: string): Promise<void> {
-      return state.changePassword(password)
+      return state.changePassword(password).then(() => {})
```
paullinator: What does this anonymous function do?

src/modules/account/account.test.js
```diff
@@ -14,8 +14,8 @@ const contextOptions = {
   plugins: [makeFakeCurrency()]
 }
 
-function findKeys (keyInfos, type) {
-  return keyInfos.find(info => info.type === type)
+function findWallet (walletInfos, type) {
```
paullinator: 😀👍

src/modules/login/login.js
swansontec: No, it's just typing. `children` can't be null in this context, but Flow doesn't know that, so I had to put in the check for the types to work. The fallback case should never be hit.

src/modules/login/login.js
swansontec: No. In this case, Flow is happy with things. `LoginStash.children` is allowed to be null.

src/modules/account/accountApi.js
swansontec: It returns `undefined`, which is what Flow needs us to return.

swansontec:
I'm tired of being blocked on this. Merging anyhow.

PR #89, opened 2018-06-05
Implement the GUI plugin data API

swansontec:


swansontec:
@wolverineks We have `deleteItem` but `clearPlugin`. Should they both be `deleteItem` & `deletePlugin`, or maybe both `clearItem` & `clearPlugin`?

PR #90, opened 2018-06-08
Shapeshift support for monero and ripple

ahujsak:


src/edge-core-index.js
```diff
@@ -543,6 +543,7 @@ export type EdgeSpendTarget = {
   publicAddress?: string,
   nativeAmount?: string,
   destMetadata?: EdgeMetadata,
+  sAddress?: string,
```
paullinator: This shouldn't be needed here. Only `otherParams.uniqueIdentifier`

src/modules/currency/wallet/currency-wallet-api.js
```diff
@@ -356,10 +356,30 @@ export function makeCurrencyWalletApi (
...
+        destTag = splitArray[1]
+      }
+
       const spendTarget: EdgeSpendTarget = {
         nativeAmount: nativeAmountForSpend,
         publicAddress: exchangeData.deposit
       }
+      if (hasDestTag) {
+        spendTarget.otherParams = {
+          uniqueIndentifier: Number(destTag)
```
paullinator: This needs to be a string. The plugin turns this into a number itself

src/edge-core-index.js
ahujsak: actually I need it to pass flow, because it comes back from shapeshift and is part of the object that is returned, so when I assign uniqueIdentifier to returnedData.sAddress if it is not defined here it does not pass flow

src/modules/currency/wallet/currency-wallet-api.js
ahujsak: nope, passing it as a string, the plug in throws an error.

ahujsak: unless you updated the plugin since I wrote this piece

PR #100, opened 2018-07-10
Typo

thehobbit85:
`paymentProtocolUrl` should be `paymentProtocolURL` according to both the specs and the actual usage.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "edge-core-js",
-  "version": "0.9.7",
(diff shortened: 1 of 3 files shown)


wolverineks:
According to Edge Conventions, the spec is wrong. Acronyms used in variable names are not all caps, they are camelCase
https://github.com/Airbitz/edge-conventions/blob/master/naming.md

swansontec:
Agreed with @wolverineks. The current version in the core is correct.

PR #114, opened 2018-08-20
Cleanups

swansontec:
Miscellaneous code janitor things.


package.json
```diff
@@ -32,7 +32,7 @@
   "module": "lib/edge-core-index.es.js",
   "repository": {
     "type": "git",
-    "url": "git@github.com:Airbitz/airbitz-core-js.git"
```
swansontec: EdgeApp

src/modules/exchange/exchange-selectors.js
```diff
@@ -1,17 +1,42 @@
 // @flow
 
 import type { RootState } from '../root-reducer.js'
+import type { ExchangePair, ExchangeRoutes } from './exchange-reducer'
```
swansontec: .js

PR #118, opened 2018-08-30
William/more cleanups

swansontec:
These are more problems I found while working on the giant account refactor.


src/modules/login/login-types.js
```diff
@@ -106,21 +106,21 @@ export type LoginTree = {
...
 // Helper for defining specific key types.
 // Use `EdgeWalletInfo` for generic wallet infos:
-interface WalletInfo<K = {}> {
-  type: string;
-  id: string;
-  keys: K;
+type WalletInfo<K = {}> = {
+  type: string,
+  id: string,
+  keys: K
```
kylanhurt: Consider using longer name

PR #119, opened 2018-08-30
William/account refactor

swansontec:
This moves the account state into Redux, and uses redux-keto to keep the `loginTree` up to date. This change is needed before the bridge can work correctly.


swansontec:
Fixed all issues. I'll wait on merging this until we finish the current release.

PR #128, opened 2018-09-24
Work around some weird Flow bug in the GUI

swansontec:


swansontec:
Verbal approval from @kylanhurt

PR #134, opened 2018-10-08
William/new exchange api

swansontec:
This is the new API Allen & I are working on.

### DIFF
--- a/src/index.js
+++ b/src/index.js
@@ -367,6 +367,15 @@ export type EdgeCurrencyToolsMap = {
   [pluginName: string]: EdgeCurrencyTools
 }
 
+export type EdgeExchangeTools = {
+  +watch: Subscriber<EdgeExchangeTools>,
+
+  +exchangeInfo: Object,
+  +settings: Object,
+
+  changeSettings(settings: Object): Promise<mixed>
+}
+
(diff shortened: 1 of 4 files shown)


src/modules/account/currency-api.js
```diff
@@ -50,3 +51,28 @@ export class CurrencyTools extends Bridgeable<EdgeCurrencyTools> {
...
+
+export class ExchangeTools extends Bridgeable<EdgeExchangeTools> {
+  _settings: Object
+
+  constructor () {
+    super()
+    this._settings = {}
+  }
+
+  get exchangeInfo () {
```
ahujsak: Is this going to barf since there is no return type.. I think this was part of what I got back today as a flow error.

src/index.js
```diff
@@ -404,6 +413,35 @@ export type EdgePluginData = {
...
+}
+
+export type EdgeExchangeQuote = {
+  +expirationDate: Date,
+  +fromNativeAmount: string,
+  +toNativeAmount: string,
+
+  +exchangeService: string,
+  +quoteId: string,
+  +quoteUri: string,
```
swansontec: TODO: Put this into `orderUri` in the metadata on approval, and remove it from the quote.

PR #142, opened 2018-10-25
fixed throw token needed filter

ahujsak:


paullinator:
Reviewed by @swansontec

PR #154, opened 2018-11-12
Fix missing transactions on calls to getTransactions

paullinator:
Change getTransactions to dispatch it's own GOT_TXS action once it has queried the plugin. Otherwise an incoming network tx would cause onTransactionsChanged to set gotTxs = true and getTransactions wouldn't further call the plugin->getTransactions when needed.


paullinator:
approved by @thehobbit85

PR #157, opened 2018-11-15
William/swap errors

swansontec:
This lays the groundwork for a better user experience with weird swap corner cases (like Shapeshift in New York).


swansontec:
Visual PR with Paul.

PR #158, opened 2018-11-18
Fix swap errors

paullinator:
* On errors throw the error directly and do not create a new error object. Creating a new error object will remove the error parameters
* Change detection of ShapeShift error for unverified user


paullinator:
@swansontec Please do a post merge review of this commit. This is getting merged now so we can get a build into review in app store. App store publish will wait until this PR is reviewed.

PR #161, opened 2018-11-20
William/remove deprecated

swansontec:
Once we get the BCH point release out, I'd like to do a breaking core version bump to include the Disklet stuff and these deprecation removals.


swansontec:
Ready for review.

PR #162, opened 2018-11-20
William/upgrade disklet

swansontec:
Once we get the BCH point release out, I would like to start using the new Disklet API.


swansontec:
Ready for review.

PR #168, opened 2018-11-25
Do not await on `reloadPluginSettings`

paullinator:
This prevents network roundtrips from blocking login. Upwards of 30s saved on really slow networks.


src/modules/account/account-pixie.js
```diff
@@ -76,10 +76,8 @@ const accountPixie = combinePixies({
         onLoggedOut = callbacks.onLoggedOut
 
         const loadAllFiles = async () => {
-          await Promise.all([
-            reloadPluginSettings(ai, accountId),
-            loadAllWalletStates(ai, accountId)
-          ])
+          reloadPluginSettings(ai, accountId)
```
swansontec: This is too high-level. We should do the change inside `reloadPluginSettings`, at the point where it actually triggers the plugins to update.

swansontec: I think we can safely run `updatePluginUserSettings` in the background. Also, removing the `await` is not enough. You need to do a `.catch()` to avoid unhandled promise rejections.

swansontec: Actually, even more specifically, within `updatePluginUserSettings`, there is a `promises.push(promise)` line right after updating the currency plugin settings that we can just delete. The `catch` is already included in that function.

PR #170, opened 2018-11-28
Fix missing txs

adyshimony:
The logic of txs slicing and search them in files didn't handle the case that a tx dropped. 

Now we make sure that the total txs number is equal to the number the ui requested. 
So after slice, verify that all txs in files, if some are dropped and missing, do it again recursively until returning the requested amount of txs.


adyshimony:
> Please no merge commits. This should be rebased on the latest develop and force-pushed to update the PR.

I did interactive rebase, hope this is what you wanted

adyshimony:
A new clean pr created

swansontec:
Thanks, the rebase looks correct.

PR #172, opened 2018-12-02
Fix missing txs - refactor getTransactions

adyshimony:
The logic of txs slicing and search them in files didn't handle the case that a tx dropped.

Now we make sure that the total txs number is equal to the number the ui requested.
So after slice, verify that all txs in files, if some are dropped and missing, do it again recursively until returning the requested amount of txs.

### DIFF
--- a/src/modules/currency/wallet/currency-wallet-api.js
+++ b/src/modules/currency/wallet/currency-wallet-api.js
@@ -253,10 +253,11 @@ export function makeCurrencyWalletApi (
       }
       // we need to make sure that after slicing, the total txs number is equal to opts.startEntries
       // slice, verify txs in files, if some are dropped and missing, do it again recursively
-      const getBulkTx = async (index: number, entries: number, out: any) => {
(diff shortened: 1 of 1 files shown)


src/modules/currency/wallet/currency-wallet-api.js
```diff
@@ -251,30 +251,39 @@ export function makeCurrencyWalletApi (
...
+            (!tx.nativeAmount[currencyCode] && !tx.networkFee[currencyCode])
+          ) {
+            continue
+          }
+          out.push(combineTxWithFile(input, tx, file, currencyCode))
         }
-
-        out.push(combineTxWithFile(input, tx, file, currencyCode))
+        // continue until the required tx number loaded
+        const res = await getBulkTx(index + entries, entries - out.length, out)
```
thehobbit85: I'm pretty sure it should be 
`startEntries - out.length` 
instead of 
`entries - out.length`

PR #177, opened 2018-12-08
Allen/change now plugin

ahujsak:


src/modules/swap/changenow-plugin.js
```diff
@@ -207,7 +192,12 @@ function makeChangeNowTools (env): EdgeSwapTools {
                 const quoteReply = await get(estQuery)
                 console.log('CN: Quote Reply', quoteReply)
                 if (quoteReply.error === 'out_of_range') {
-                  throw new SwapOutOfRangeError(swapInfo, nativeMin, nativeMax)
+                  if (lt(quoteParams.from, item.min.toString())) {
+                    throw new SwapBelowLimitError(swapInfo, nativeMin)
+                  }
+                  if (gt(quoteParams.from, item.max.toString())) {
+                    throw new SwapAboveLimitError(swapInfo, nativeMax)
+                  }
```
swansontec: This should just be an `else` case.

PR #191, opened 2019-01-20
Fix getTransactions for tokens in accountbased

paullinator:
state.gotTxs was a single boolean for all tokens in a wallet. This could cause only a single token's transactions to get read from the currency plugin. If a future call to getTransactions is called for a different token, Core won't call getTransactions for that token. This bug was hidden because only the ethereum plugin had tokens and it would always update the Core with all transactions upon boot. accountbased is optimized to only load transactions once getTransactions is called.

Fix is to make state.gotTxs a boolean map by currencyCode


swansontec:
Just waiting for Travis now.

PR #195, opened 2019-01-29
Add an `EdgeContext.close` method

swansontec:
As it says on the tin.


src/modules/account/account-reducer.js
```diff
@@ -278,7 +278,11 @@ const account = buildReducer({
 export const accountReducer = filterReducer(
   account,
   (action: RootAction, next: AccountNext) => {
-    if (/^ACCOUNT_/.test(action.type) && action.payload.accountId === next.id) {
+    if (
+      /^ACCOUNT_/.test(action.type) &&
+      action.payload &&
```
swansontec: Should be `action.payload != null`. Oops.

PR #198, opened 2019-02-04
William/combine plugins

swansontec:
This is a non-breaking change. It moves the plugin management into a central location to reduce code duplication.

### DIFF
--- a/src/modules/plugins/plugins-actions.js
+++ b/src/modules/plugins/plugins-actions.js
@@ -1,0 +1,53 @@
+// @flow
+
+import { downgradeDisklet } from 'disklet'
+import { type Dispatch } from 'redux'
+
+import {
+  type EdgeCorePlugin,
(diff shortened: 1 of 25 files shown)


src/modules/plugins/plugins-actions.js
```diff
@@ -0,0 +1,53 @@
...
+  } catch (error) {
+    dispatch({ type: 'CORE_PLUGINS_FAILED', payload: error })
+  }
+
+  // Add the built-in swap plugins:
+  payload.changelly = changellyPlugin
+  payload.changenow = changenowPlugin
+  payload.faast = faastPlugin
+  payload.shapeshift = shapeshiftPlugin
+
```
kylanhurt: Does HERC need to be added to this list now?

swansontec: I don't think so, since that plugin lives in another repo.

PR #213, opened 2019-03-05
Update index.js.flow

wolverineks:


wolverineks:
Not sure if this file is created automatically, but it looks like the new core plugin functions should be here to make flow happy

index.js.flow
```diff
@@ -12,6 +12,8 @@ import {
 
 export * from './src/types/types.js'
 
+declare export function addEdgeCorePlugins(): mixed
```
swansontec: Good catch, but we need to set up the parameter type here.

swansontec:
Cherry-picked, adjusted, & merged.

swansontec:
Published in v0.15.4.

PR #216, opened 2019-03-18
Throttle callbacks on a global level

paullinator:
This ensures that multiple wallets do not hammer the GUI with simultaneous callbacks, especially at startup.

### DIFF
--- a/src/util/updateQueue.js
+++ b/src/util/updateQueue.js
@@ -1,0 +1,62 @@
+// @flow
+
+// How often to run jobs from the queue
+let QUEUE_RUN_DELAY = 500
+
+// How many jobs to run from the queue on each cycle
+let QUEUE_JOBS_PER_RUN = 3
+
+type UpdateQueue = {
+  id: string,
+  action: string,
+  updateFunc: Function
+}
+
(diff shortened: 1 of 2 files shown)


src/util/updateQueue.js
```diff
@@ -0,0 +1,64 @@
...
+  setTimeout(() => {
+    const numJobs =
+      QUEUE_JOBS_PER_RUN < updateQueue.length
+        ? QUEUE_JOBS_PER_RUN
+        : updateQueue.length
+    if (numJobs) console.log('Core Queue running more jobs...')
+    for (let i = 0; i < numJobs; i++) {
+      if (updateQueue.length) {
+        const u = updateQueue.shift()
+        console.log(`Core Job: ${u.id} ${u.action}`)
```
swansontec: This logging makes me sad. I should be able to run the unit tests, and only see green check marks, not random debug stuff interleaved. If it's important enough to log, use the `io.console.log`. If it's not important enough to log, please remove it.

paullinator: Good catch. The logging wasn't meant to be there. forcepushed the remove of those. Thx.

PR #219, opened 2019-03-23
Cache the public keys per wallet

swansontec:


swansontec:
Tested on my phone and on the iOS simulator - we are no longer calling `derivePublicKeys` on subsequent loads.

PR #237, opened 2019-07-20
Add otherParams (from currencyPlugin) to EdgeTransaction

kylanhurt:
Allow currency-plugins to add data to `otherParam` that pass through to core (and GUI), specifically for debugging info (inputs, outputs, satoshis, etc)


swansontec:
Tests failing due to `yarn audit`, which is fixed separately. We should really fix the Travis script not to care about this, now that GitHub also includes this feature natively.

PR #239, opened 2019-07-31
Allow importPrivateKey to be sync

kylanhurt:


swansontec:
Our long-term goal is to make all plugin methods async, so we can run plugins inside Web Workers or something like that.

PR #245, opened 2019-08-10
William/pausing

swansontec:
This seems to work locally, but obviously will require a lot of testing.


src/core/currency/wallet/currency-wallet-pixie.js
```diff
@@ -209,39 +175,95 @@ export const walletPixie: TamePixie<CurrencyWalletProps> = combinePixies({
...
+        destroy() {
+          const { id } = input.props
+          if (!input.props.selfOutput) return
+
+          const { engine } = input.props.selfOutput
+          if (engine != null && started) {
+            started = false
+            input.props.io.console.info(`${id} killEngine`)
+            try {
+              engine.killEngine()
```
thehobbit85: Set flag on engine to know if it's active or not

PR #247, opened 2019-08-20
Fix the account plugin yaob sync

swansontec:
Some plugin property changes weren't being synced up to the GUI when the files changed in git. Now they are.


swansontec:

PR #248, opened 2019-08-24
Close unused swap quotes

swansontec:


swansontec:

PR #251, opened 2019-09-23
Fix the swap quote prioritization logic

swansontec:
We would sometimes return estimates instead of accurate quotes, due to a hole in the logic.


swansontec:

PR #267, opened 2019-12-18
EF-991 add hidden EdgeWalletState

bruscantini:
Add "hidden" wallet state value. Will be used by react-gui to hide certain wallets. Needed to hide user wallets created by Ember from Edge App.

### DIFF
--- a/src/core/account/account-files.js
+++ b/src/core/account/account-files.js
@@ -75,7 +75,7 @@ function loadWalletList(
     const walletStates = {}
 
     files.forEach(file => {
-      const { SortIndex, Archived, BitcoinSeed, MK, SyncKey } = file.json
+      const { SortIndex, Archived, Hidden, BitcoinSeed, MK, SyncKey } = file.json
 
       const dataKey = base16.parse(MK)
       const bitcoinKey = base16.parse(BitcoinSeed)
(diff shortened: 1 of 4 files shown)


src/core/account/account-files.js
```diff
@@ -75,7 +75,7 @@ function loadWalletList(
     const walletStates = {}
 
     files.forEach(file => {
-      const { SortIndex, Archived, BitcoinSeed, MK, SyncKey } = file.json
+      const { SortIndex, Archived, Hidden, BitcoinSeed, MK, SyncKey } = file.json
```
swansontec: This is the legacy Airbitz wallet list, which won't ever have `Hidden` support, since that app is deprecated. We should just default `hidden` to `false` in this routine.

src/types/types.js
```diff
@@ -120,7 +121,8 @@ export type EdgeWalletInfoFull = {
 export type EdgeWalletState = {
   archived?: boolean,
   deleted?: boolean,
-  sortIndex?: number
+  sortIndex?: number,
+  hidden?: boolean
```
swansontec: Can you alphabetize this?

swansontec:
I went ahead and cherry-picked this by hand, along with the changes I requested.

bruscantini:
Thanks for the speedy review @swansontec !
I opened the edge-react-gui PR here:
https://github.com/EdgeApp/edge-react-gui/pull/1761

PR #268, opened 2019-12-23
William/logging

swansontec:
We wanted this years ago, and I've slowly been chipping away at it. Now it's ready.

### DIFF
--- a/src/core/log/log.js
+++ b/src/core/log/log.js
@@ -1,0 +1,45 @@
+// @flow
+
+import {
+  type EdgeIo,
+  type EdgeLog,
+  type EdgeLogMethod
+} from '../../types/types.js'
+import { addHiddenProperties } from '../../util/util.js'
+
+type EdgeLogType = 'info' | 'warn' | 'error'
+
+type EdgeLogEvent = {
+  type: EdgeLogType,
+  time: Date,
+  sender: string,
+  message: string
+}
+
+function makeLogMethod(
+  io: EdgeIo,
+  type: EdgeLogType,
+  sender: string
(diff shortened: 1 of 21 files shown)


src/core/log/log.js
```diff
@@ -0,0 +1,42 @@
...
+  return function log() {
+    let message = ''
+    for (let i = 0; i < arguments.length; ++i) {
+      const arg = arguments[i]
+      if (i > 0) message += ' '
+      message += typeof arg === 'string' ? arg : JSON.stringify(arg)
+    }
+
+    const entry: EdgeLogEntry = { type, time: new Date(), sender, message }
+    io.console.info(entry.sender + ': ' + entry.message)
```
swansontec: I have more commits in this branch, which add an onLog event within the `makeEdgeContext` setup. However, that stuff isn't ready yet. In the mean time, this `EdgeLogEntry` is the format of the event.

src/core/log/log.js
```diff
@@ -0,0 +1,42 @@
...
+  io: EdgeIo,
+  type: EdgeLogType,
+  sender: string
+): EdgeLogMethod {
+  return function log() {
+    let message = ''
+    for (let i = 0; i < arguments.length; ++i) {
+      const arg = arguments[i]
+      if (i > 0) message += ' '
+      message += typeof arg === 'string' ? arg : JSON.stringify(arg)
```
paullinator: pretty?

src/core/log/log.js
paullinator: time?

PR #272, opened 2020-01-16
Log auth server failures

swansontec:
This will give us more details in the logs if a "Could not reach the auth server" error happens.


swansontec:
Example log output: `01-15 16:13:24.789: I/ReactNativeJS(21275): 01-16 00:13:24 edge-core: POST https://auth.airbitz.co/api/v2/login failed in 14ms, TypeError: Failed to fetch`

PR #273, opened 2020-01-20
William/more fixes

swansontec:
I copied a bunch of stuff from the core to the login server, and in doing so, realized they needed to be fixed in various ways.

The biggest commit is the last one, which accurately re-creates the Python status codes in Javascript. I've been cribbing off of this commit pretty heavily in the login server work.

The first commit, folding `fake-socket.js` into `fake-io.js`, means the core now has Flow activated on every source file (took a few years).


swansontec:
Verbal PR approval with @paullinator.

PR #276, opened 2020-01-31
William/swap plugins

swansontec:
This lets us have preferred swap providers.

Also, we've been talking about `pluginId` in our designs for the feature, but the core has been using `pluginName`, which is confusing, and has lead to problems like people trying to show the `pluginName` to the user. Using the "id" terminology should help put a stop to that.


swansontec:
Got verbal approval from @paullinator last week.

PR #281, opened 2020-03-04
William/type cleanups

swansontec:
Here are some things from the Typescript branch.

### DIFF
--- a/src/core/swap/swap-api.js
+++ b/src/core/swap/swap-api.js
@@ -132,14 +132,15 @@ function pickBestError(errors: any[]): any {
 /**
  * Ranks different error codes by priority.
  */
-function rankError(error: any) {
-  if (error.name === errorNames.InsufficientFundsError) return 5
-  if (error.name === errorNames.PendingFundsError) return 5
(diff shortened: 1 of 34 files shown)


src/core/swap/swap-api.js
```diff
@@ -132,7 +132,8 @@ function pickBestError(errors: any[]): any {
 /**
  * Ranks different error codes by priority.
  */
-function rankError(error: any) {
+function rankError(error: any): number {
+  if (error == null) return 0
```
swansontec: -1

src/util/promise.js
```diff
@@ -25,12 +24,19 @@ export function timeout<T>(
...
-    const onDone = () => clearTimeout(timer)
-    promise.then(onDone, onDone)
+  return new Promise((resolve, reject) => {
+    const timer = setTimeout(() => reject(error), ms)
+    promise.then(
+      ok => {
+        resolve(ok)
+        clearTimeout(timer)
+      },
+      bad => {
```
swansontec: s/bad/error/

PR #290, opened 2020-05-15
Add special metadata for swaps

swansontec:
# Swap Metadata

Since the Airbitz days, the `EdgeMetatada` structure has only contained user-editable information. Things the user can't edit, like the first-seen date or partner revenue accounting, go directly into the transaction object, not `EdgeMetadata`. This is because we don't want the `saveTxMetadata` function to overwrite these fields.

So, the swap information needs to go somewhere on the transaction other than the `EdgeMetadata` structure. I propose creating a new data structure that exchanges pass as `EdgeSpendInfo.swapData`. The core saves this structure, and then puts it on the `EdgeTransaction` when available. When the transaction details scene sees that `edgeTransaction.swapData != null`, it can display the transaction as a swap and add the extra elements to the UI.

```typescript
type EdgeTxSwap = {
  orderId: string
  orderUri: string
  isEstimate: boolean

  // The EdgeSwapInfo from the swap provider:
  plugin: {
    pluginId: string
    displayName: string
    supportEmail?: string
  }

  // Address information:
  payoutAddress: string
  payoutCurrencyCode: string
  payoutNativeAmount: string
  payoutWalletId: string
  refundAddress: string
}
```

This structure does not contain the exchange receive address. This is relevant to all transactions, not just swaps. Whenever the user sends money, the core should save the `EdgeSpendTarget` fields as `EdgeTransaction.spendTargets`. This would help privacy coins like Monero remember basic information, it would help coins with change remember which addresses were the true destination, and it would help us debug things like the Tezos wrong addresses. The transaction details screen could display this for all transactions that have it, including swaps.

## Original specification

The original specification called for this information to live in `metadata.otherParams.exchange`, but metadata is the wrong location. The new locations is `EdgeTransaction.swapData` & `EdgeSpendInfo.swapData`, as explained above.

The metadata is supposed to contain a `status` property, but there is no information about how to keep this updated. Keeping a background process running to update the metadata seems inefficient. Instead, we could implement a `checkStatus(orderId: string)` method for the transaction details screen to call as needed.

The other fields in the specification have straightforward mappings to the new design:

- status -> `account.swapConfig[pluginId].checkStatus(orderId)`
- orderId -> `tx.swapData.orderId`
- statusUrl -> `tx.swapData.orderUri` (name used in EdgeSwapQuote)
- pluginId -> `tx.swapData.plugin.pluginId`
- supportEmail -> `tx.swapData.plugin.supportEmail`
- pluginName -> `tx.swapData.plugin.displayName`
- sourceCurrencyCode -> `tx.spendTargets[0].currencyCode`
- sourceAmount -> `tx.spendTargets[0].nativeAmount`
- destinationWalletName -> `tx.swapData.payoutWalletId`
- destinationCurrencyCode -> `tx.swapData.payoutCurrencyCode`
- destinationAmount -> `tx.swapData.payoutNativeAmount`
- receiveAddress -> `tx.swapData.payoutAddress`
- refundAddress -> `tx.swapData.refundAddress`
- quoteType -> `tx.swapData.isEstimate`

Some names are different to match what the `EdgeSwapQuote` does, while "destination" and "receive" names now use "payout" to clear up the ambiguity between the exchange or the final wallet.

## This PR

This PR provides everything above except the `checkStatus` method.

### DIFF
--- a/src/core/currency/wallet/currency-wallet-callbacks.js
+++ b/src/core/currency/wallet/currency-wallet-callbacks.js
@@ -159,44 +159,51 @@ export function makeCurrencyWalletCallbacks(
           return
         }
       }
+
+      // Grab stuff from redux:
       const { state } = input.props
-      const existingTxs = input.props.selfState.txs
-      const txidHashes = {}
-      const files = input.props.selfState.files || {}
-      const fileNames = input.props.selfState.fileNames || []
+      const {
+        fileNames,
+        fileNamesLoaded,
+        txs: reduxTxs
+      } = input.props.selfState
       const defaultCurrency = input.props.selfState.currencyInfo.currencyCode
+
+      const txidHashes = {}
       const changed = []
       const created = []
-      for (const rawTx of txs) {
-        const tx = mergeTx(rawTx, defaultCurrency, existingTxs[rawTx.txid])
-        const txid = tx.txid
-        // If we already have it in the list, make sure something about it has changed:
-        if (compare(tx, existingTxs[txid])) continue
+      for (const tx of txs) {
+        const { txid } = tx
 
+        // Verify that something has changed:
+        const reduxTx = mergeTx(tx, defaultCurrency, reduxTxs[txid])
+        if (compare(reduxTx, reduxTxs[txid])) continue
+
+        // Ensure the transaction has metadata:
         const txidHash = hashStorageWalletFilename(state, walletId, txid)
-        const isNew = !fileNames[txidHash]
-        const decryptedMetadata = files[txidHash]
-        const combinedTx = combineTxWithFile(
-          input,
-          tx,
-          decryptedMetadata,
-          rawTx.currencyCode
-        )
+        const isNew = fileNamesLoaded && fileNames[txidHash] == null
         if (isNew) {
           setupNewTxMetadata(input, tx).catch(e => input.props.onError(e))
-          created.push(combinedTx)
-        } else if (decryptedMetadata) {
-          changed.push(combinedTx)
         }
+
+        // Build the final transaction to show the user:
+        const { files } = input.props.selfState
+        const combinedTx = combineTxWithFile(
+          input,
+          reduxTx,
+          files[txidHash],
+          tx.currencyCode
+        )
+        if (isNew) created.push(combinedTx)
+        else if (files[txidHash] != null) changed.push(combinedTx)
         txidHashes[txidHash] = combinedTx.date
       }
-      // Side Effect
+
+      // Tell everyone who cares:
       input.props.dispatch({
         type: 'CURRENCY_ENGINE_CHANGED_TXS',
         payload: { txs, walletId, txidHashes }
       })
-
-      // Call the callbacks:
       if (changed.length) throtteldOnTxChanged(changed)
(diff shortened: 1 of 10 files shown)


test/core/currency/wallet/currency-wallet.test.js
```diff
@@ -195,25 +196,49 @@ describe('currency wallets', function () {
...
+    await config.changeUserSettings({ txs: { a: { nativeAmount: '25' } } })
+    await wallet.saveTx(tx)
+
+    const txs = await wallet.getTransactions({})
+    expect(txs.length).equals(1)
+    expect(txs[0].nativeAmount).equals('25')
+
+    const { metadata } = tx
+    if (metadata == null) throw new Error('No metadata')
+    expect(metadata.name).equals(metadata.name)
```
peachbits, line 212: Shouldn't this be 

`expect(metadata.name).equals('me')`

src/types/types.js
```diff
@@ -265,6 +265,13 @@ export type EdgeMetadata = {
   notes?: string,
   amountFiat?: number,
   bizId?: number,
+  swap?: {
+    +pluginId: string,
+    +orderId: string,
+    +orderUri?: string,
+    +payoutAddress: string,
+    +payoutCurrencyCode: string
+  },
```
peachbits, line 274: Missing a lot of parameters and this also changed the naming proposed in the design doc. These were the following parameters for swap:

orderId
statusUrl
supportEmail
pluginId
pluginName
sourceCurrencyCode
sourceAmount (in units of exchange amount)
destinationWalletName
destinationCurrencyCode
destinationAmount (in units of exchange amount)
receiveAddress (this is user address used to receive exchange funds)
refundAddress
quoteType: 'fixed' | 'variable'

src/types/types.js
```diff
@@ -265,6 +265,13 @@ export type EdgeMetadata = {
   notes?: string,
```
peachbits, line 265: This is missing recipientAddress param which will be used for all outgoing transaction including for swaps.

test/core/currency/wallet/currency-wallet.test.js
swansontec: Fixed.

src/types/types.js
swansontec: We have a new design document that this follows.

src/types/types.js
swansontec: This is handled a different way in the new design document.

src/core/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -159,44 +159,51 @@ export function makeCurrencyWalletCallbacks(
...
-          decryptedMetadata,
-          rawTx.currencyCode
-        )
+        const isNew = fileNamesLoaded && fileNames[txidHash] == null
         if (isNew) {
           setupNewTxMetadata(input, tx).catch(e => input.props.onError(e))
-          created.push(combinedTx)
-        } else if (decryptedMetadata) {
-          changed.push(combinedTx)
         }
```
swansontec, line 187: `else if (metadata doesn't have this coin) updateFiatMetadata()`

PR #292, opened 2020-06-01
Save a decryption secret with spend transactions

swansontec:
I don't remember what we agreed to call it, so I just called it "secret". We can change that name pretty easily now that it's working.


src/core/currency/wallet/currency-wallet-files.js
```diff
@@ -45,6 +45,7 @@ export type TransactionFile = {
     currency: string,
     tag?: string
   }>,
+  secret?: string,
```
paullinator, line 48: Can you change to `txSecret`. Thats what we had decided and its easier to search for.

PR #299, opened 2020-07-05
Add additional fee information to EdgeTransaction

peachbits:


paullinator:
> Storing English strings in here is a terrible design.

I agree it is not great. But since each currency might have drastically different formats for reporting transaction fee rates, and since this is mostly a troubleshooting feature, this seemed like the best option. Otherwise each currency would have to communicate the format of its rates. As well, these rates would not have a lot of English specific terminology, but mostly technical terms. ie. "1 wei gasprice, 21000 gaslimit"

swansontec:
Replaced by #304.

PR #310, opened 2020-07-20
William/login features

swansontec:


src/core/login/create.js
```diff
@@ -60,6 +60,7 @@ export function makeCreateKit(
 
   const dummyLogin: LoginTree = {
     appId,
+    created: new Date(),
```
swansontec, line 63: Could be undefined here.

src/core/login/login.js
```diff
@@ -81,6 +81,7 @@ function applyLoginReplyInner(
   // Copy common items:
   const out: LoginStash = filterObject(loginReply, [
     'appId',
+    'created',
```
swansontec, line 84: Verify that the date is in the LoginReply cleaner.

swansontec: Fixed.

src/core/login/create.js
swansontec: Fixed.

PR #316, opened 2020-09-07
Add more voucher features

swansontec:
This lets the core manage the list of pending vouchers, so it stays better in sync with the server.


swansontec (commented):
Verbal PR with @paullinator and @thehobbit85.

PR #318, opened 2020-09-16
Don't crash on barcode logins with OTP reset dates

swansontec:
We weren't cleaning the server replies properly, which meant dates were sometimes still strings.


paullinator (approved):
Approved by @thehobbit85

PR #319, opened 2020-09-17
Shred vouchers on barcode login

swansontec:


swansontec (commented):
Verbal PR with @thehobbit85.

PR #321, opened 2020-11-06
RBF Transaction Support

samholmes:
* Adds new `rbfTxid` optional string to `EdgeSpendInfo` type definition
* Adds new `rbfTxid` to the returned `EdgeTransaction` object in
`makeSpend` on `EdgeCurrencyWallet` objects


swansontec (approved):
Yep, this seems about right.

PR #324, opened 2020-11-24
Filter getTransactions by searchString

peachbits:


src/core/currency/wallet/currency-wallet-export.js
```diff
@@ -24,6 +24,48 @@ export function dateFilter(
   return out
 }
 
+export function searchStringFilter(
+  txs: EdgeTransaction[],
+  opts: EdgeGetTransactionsOptions
+): EdgeTransaction[] {
+  const { searchString } = opts
+
+  let out = txs
```
paullinator, line 33: normalize all search strings and transaction information to lowercase Before search to make it case insensitive

paullinator (changes requested):

paullinator (approved):

PR #331, opened 2020-12-22
Edge Parse Uri options type.

andreyvEze:


src/core/currency/wallet/currency-wallet-api.js
```diff
@@ -17,6 +17,7 @@ import {
   type EdgeGetTransactionsOptions,
   type EdgeMetadata,
   type EdgeParsedUri,
+  type EdgeParseUriOptions,
```
paullinator, line 20: Make this a cleaner

src/types/types.js
```diff
@@ -524,6 +524,12 @@ export type EdgeCurrencyEngine = {
 
 // currency plugin -----------------------------------------------------
 
+export type EdgeParseUriOptions = {
```
paullinator, line 527: Make this a cleaner and export it for use by the plug-ins. It will clean up the code that tries to validate the options

src/core/currency/wallet/currency-wallet-api.js
swansontec: Per our discussion, a cleaner is not needed. The other code should simply trust the Flow / Typescript system to do it's job, and not validate stuff it doesn't need to.

PR #335, opened 2021-01-11
Overwrite publicAddress when encoding URI

passabilities:
The type `EdgeEncodeUri` should not contain `segwitAddress` or `legacyAddress` as it should only be dealing with 1 address (the `publicAddress`)

### DIFF
--- a/src/types/types.js
+++ b/src/types/types.js
@@ -438,8 +438,8 @@ export type EdgeParsedUri = {
 
 export type EdgeEncodeUri = {
   publicAddress: string,
-  segwitAddress?: string,
-  legacyAddress?: string,
+  segwitAddress?: string, // Deprecated. Use publicAddress instead.
+  legacyAddress?: string, // Deprecated. Use publicAddress instead.
   nativeAmount?: string,
   label?: string,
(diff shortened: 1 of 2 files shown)


src/types/types.js
```diff
@@ -431,8 +431,6 @@ export type EdgeParsedUri = {
 
 export type EdgeEncodeUri = {
   publicAddress: string,
-  segwitAddress?: string,
-  legacyAddress?: string,
```
swansontec, line 435: The `encodeUri` method in the core should just take whatever is passed in either of these fields & copy it to `publicAddress`, so plugins can simply ignore these other two & just read `publicAddress`.

Instead of deleting these two from the public Flow types, but instead simply mark them as `// Deprecated`, until our next major version bump, so that this isn't a breaking change.

swansontec (commented):
This would be a breaking change, so we need to do it a bit differently.

swansontec (commented):
I have updated the PR based on my newest comment, @passabilities. Feel free to press the "merge" button if the updates look OK.

swansontec (approved):

PR #336, opened 2021-01-14
Log EdgeSwapRequest

peachbits:


paullinator (changes requested):
This is missing the logging that shows the results of each swap provider. That was the purpose of the logging initially added to each of the exchange plug-ins. The goal is to see what error and or successful quote each provider is returning, Regardless of which one was chosen.

src/core/swap/swap-api.js
```diff
@@ -25,13 +25,12 @@ export async function fetchSwapQuote(
...
   const { log } = ai.props
 
-  log.warn(
-    `EdgeSwapRequest fromCurrencyCode ${request.fromCurrencyCode} toCurrencyCode ${request.toCurrencyCode} nativeAmount ${request.nativeAmount} quoteFor ${request.quoteFor}`
-  )
   const account = ai.props.state.accounts[accountId]
   const { swapSettings, userSettings } = account
   const swapPlugins = ai.props.state.plugins.swap
 
+  console.warn('Requesting swap quotes for: ', request)
```
peachbits, line 32: ???

swansontec: OOops, should be `log.warn`. Told ya it was untested.

PR #337, opened 2021-01-22
Periodically perform a re-login to sync the account

swansontec:
This will keep the account credentials in sync, including wallets and pending login requests.


swansontec:
Please hold off on merging - I found a bug. Sync failures will block git syncing, so I need a try/catch.

samholmes (approved):

PR #338, opened 2021-01-26
Make the currency engine's address methods async

swansontec:
This is not a breaking change, since non-async return values are still acceptable.

This has not been tested.


passabilities (approved):
Tested 👍🏼

PR #342, opened 2021-02-01
Make sure exchange rate routes eventually point to a fiat currency

peachbits:


src/core/exchange/exchange-selectors.js
```diff
@@ -102,6 +102,16 @@ export function getExchangeRate(
 
   // Only search if the endpoints exist:
   if (search.routes[fromCurrency] && search.routes[toCurrency]) {
+    // Make sure that routes eventually point to a fiat currency
+    if (!/^iso:/.test(fromCurrency)) {
+      for (const fromRoute in search.routes[fromCurrency]) {
```
swansontec, line 107: Please use `for .. of Object.keys(x)` instead of `for .. in x`.

src/core/exchange/exchange-selectors.js
```diff
@@ -102,6 +102,16 @@ export function getExchangeRate(
...
+        for (const deepRoute in search.routes[fromRoute]) {
```
swansontec, line 108: This second loop seems to have no effect? It just loops more or less times, and then we move to the add hint line in any case.

Suggestion: move the two loops into a function that returns a boolean, and base `addHint` on the result of calling the function. Then you can bail out of the middle loop with a `return true` (or false?) statement.

swansontec (changes requested):

src/core/exchange/exchange-selectors.js
```diff
@@ -102,6 +118,11 @@ export function getExchangeRate(
 
   // Only search if the endpoints exist:
   if (search.routes[fromCurrency] && search.routes[toCurrency]) {
+    // Make sure that routes eventually point to a fiat currency
+    if (!hasDeepFiatRoute(search, fromCurrency))
```
swansontec, line 122: I feel like the previous version had another `!isFiatCode(fromCurrency)`, which is now missing. Maybe that check belongs in the `hasDeepFiatRoute` method as a quick pre-check before doing the loop?

src/core/exchange/exchange-selectors.js
```diff
@@ -102,6 +118,11 @@ export function getExchangeRate(
...
+      for (const route of Object.keys(search.routes[fromCurrency])) {
```
swansontec, line 123: I don't understand the need for this for loop, but I guess it works?

swansontec (approved):
I see two things that are concerning, but I don't understand the logic enough to know if they are bugs are not.

So, I'm approving. Feel free to rebase & merge, unless my questions really do point to bugs.

src/core/exchange/exchange-selectors.js
```diff
@@ -78,6 +78,31 @@ function searchRoutes(
...
+function createDeepFiatRoutes(
+  search: ExchangeSearch,
+  fromCurrency: string
+): void {
+  const newHints = []
+  for (const fromRoute of Object.keys(search.routes[fromCurrency])) {
+    for (const deepRoute of Object.keys(search.routes[fromRoute])) {
+      if (isFiatCode(deepRoute)) {
+        return
+      } else {
```
swansontec, line 98: The `return; } else {` pattern isn't needed, since the `return` statement prevents the next line from running anyhow. Therefore, we can do:

```js
if (isFiatCode(deepRoute)) return
newHints.push({ fromCurrency: fromRoute, toCurrency: 'iso:USD' })
```

Since this is just a style change, feel free to toss it in when you squash away the `fixup!` commit.

swansontec (approved):
Approved. Please rebase & merge.

PR #343, opened 2021-02-02
Refactor to cover all null cases

peachbits:
Also remove leading zeroes from searchString to handle searching for less than a whole unit since values are stores as smallest denomination with no decimal places.


swansontec (approved):

swansontec:
This got merged as part of #347

PR #347, opened 2021-02-10
Refactor getTransactions

peachbits:
-Fixes search result issues by filtering per transaction
-Add date and searchString tests
-Refactor from recursive function to while loop


swansontec (approved):
I wrote my comments in the form of two fixup commits. If you like them, just rebase onto master & merge. Everything else looks good.

PR #358, opened 2021-03-14
Verify Engine Tx Cache

passabilities:
### Transaction Cache 
After transactions are fetched from the engine, the core marks in its state that the transactions have been cached via the `state.gotTxs[currencyCode]`. However, if transactions are fetched from the engine before the entire account history has been downloaded, the core will have already marked all the transactions as cached and will not try to fetch any more from the engine. 

To get around this, I compare the count of cached transactions to the count the engine returns via `engine.getNumTransactions({ currencyCode })` function.


thehobbit85:
Not needed if plugin callbacks are firing correctly.

PR #354, opened 2021-03-02
William/cleaner fixes

swansontec:
This gets the cleaners to the point were we can pretty much copy-paste them into the login server.

PR #361, opened 2021-03-25
William/uncleaners

swansontec:
This fixes a large number of type errors exposed by working with the login server. It also fixes the failure to create account & failure log into Airbitz accounts on React Native, which were introduced in #354.

### DIFF
--- a/src/core/login/create.js
+++ b/src/core/login/create.js
@@ -1,11 +1,12 @@
 // @flow
 
+import { uncleaner } from 'cleaners'
 import { base64 } from 'rfc4648'
 
 import {
-  type CreateLoginPayload,
-  type LoginRequestPayload
-} from '../../types/server-types.js'
+  asChangeSecretPayload,
+  asCreateLoginPayload
+} from '../../types/server-cleaners.js'
 import {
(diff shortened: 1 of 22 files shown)


src/core/login/create.js
```diff
@@ -92,28 +94,34 @@ export function makeCreateKit(
...
-      server,
+      server: wasCreateLoginPayload({
+        appId,
+        loginId,
+        parentBox,
+        ...keysServer,
+        ...passwordServer,
+        ...pin2Server,
+        ...secretServer
+      }),
```
samholmes, line 124: Move the splats outside the uncleaner `wasCreateLoginPayload` for clarity

samholmes (approved):
Approved with one minor fix! 👏

src/core/login/create.js
samholmes: I'm going to retract the need for this change. On second thought, it seems to make sense that we keep it in the uncleaner input. The reason for this is that we want to allow the extra fields to pass the static type check for wasCreateLoginPayload.

samholmes (approved):

PR #363, opened 2021-03-28
Fix the corrupted vouchers

swansontec:


swansontec:
New version forthcoming with even more bug fixes.

thehobbit85 (approved):

PR #370, opened 2021-04-13
William/breaking changes

swansontec:
These will take us to the 0.18.0 release.


swansontec:
We should think about leaving on the final commit in this PR, "Always return promises from currency engine methods". If we allow the currency engines to return `T | Promise<T>`, then there is no need to do breaking changes in those repos - they can keep returning plain values as they have before, instead of being forced to used promises.

src/core/account/account-init.js
```diff
@@ -91,7 +91,7 @@ export async function makeAccount(
   appId: string,
   loginTree: LoginTree,
   loginType: LoginType,
-  opts: EdgeAccountOptions = {}
+  opts: EdgeAccountOptions
```
thehobbit85, line 94: I'm guessing that it's not optional anymore since all the places it's called already has a default value?

thehobbit85 (approved):

thehobbit85:
> We should think about leaving on the final commit in this PR, "Always return promises from currency engine methods". If we allow the currency engines to return `T | Promise<T>`, then there is no need to do breaking changes in those repos - they can keep returning plain values as they have before, instead of being forced to used promises.

I agree 100%

src/core/account/account-init.js
swansontec: It's still optional, because the parameter list has it as `opts = {}` on all the `EdgeContext` methods. The difference is that passing `null` is no longer valid. By the time we get into the heart of the core, though, it's no longer optional because the `opts = {}` bit ensures it already exists.

PR #377, opened 2021-05-18
Fix initializing rateHintCache

wolverineks:
Attempting to access a non-existent file in node throws an error that
does not expose a type


wolverineks:
ive verifiede this works on node,
i havent checked if it still works on RN

PR #382, opened 2021-06-16
Add optional amount field to insufficient funds error

peachbits:


src/types/error.js
```diff
@@ -38,8 +38,9 @@ export class DustSpendError extends Error {
 export class InsufficientFundsError extends Error {
   name: string
   +currencyCode: string | void
+  +amount: string | void
```
swansontec, line 41: I'm not sure what this means. Is it `sentAmount`, or `availableAmount`? Maybe add a comment, or improve the name.

swansontec (changes requested):

swansontec (changes requested):

swansontec (approved):

PR #385, opened 2021-06-22
Re-enable ES 2020 syntax features

swansontec:
This instructs Webpack to transform away `??` and `?.` before they become a problem.

### DIFF
--- a/src/types/error.js
+++ b/src/types/error.js
@@ -58,7 +58,7 @@ export class InsufficientFundsError extends Error {
       }
     } else {
       const { currencyCode, networkFee } = opts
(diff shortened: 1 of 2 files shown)


src/types/error.js
```diff
@@ -58,7 +58,7 @@ export class InsufficientFundsError extends Error {
       }
     } else {
       const { currencyCode, networkFee } = opts
-      super(`Insufficient ${currencyCode != null ? currencyCode : 'funds'}`)
+      super(`Insufficient ${currencyCode ?? 'funds'}`)
```
thehobbit85, line 61: You don't understand how happy this makes me :)

thehobbit85 (approved):

PR #387, opened 2021-07-30
Fix transaction export

swansontec:
Sometimes transactions from the Ethereum engine have an undefined `nativeAmount`. so just insert "0" when this happens.


swansontec:
Tested on the rj131 "green" wallet, and the error no longer appears.

thehobbit85 (approved):

PR #388, opened 2021-07-30
Use sucrase in development mode

peachbits:


webpack.config.js
```diff
@@ -24,20 +24,31 @@ module.exports = {
...
-        include: path.resolve(__dirname, 'node_modules/buffer/index.js'),
-        use: {
-          loader: 'babel-loader',
-          options: { presets: ['@babel/preset-env'] }
-        }
-      }
-    ]
+    rules: production
+      ? [
+          {
```
swansontec, line 29: Please only replace the first thing in the array. The second one needs to be enabled in both modes.

webpack.config.js
```diff
@@ -24,20 +24,31 @@ module.exports = {
...
+            test: /\.js$/,
+            exclude: /node_modules/,
+            use: { loader: 'babel-loader', options: babelOptions }
+          },
+          {
+            include: path.resolve(__dirname, 'node_modules/buffer/index.js'),
```
swansontec, line 35: This guy needs to stay.

swansontec (changes requested):
Please also change the commit message to something like "Use sucrase in webpack development mode", so we know it's about webpack.

swansontec (approved):

PR #394, opened 2021-09-09
Limit the maximum of files in changes per syncRepo request

samholmes:
This should alleviate max payload errors encountered by the server.

### DIFF
--- a/src/core/storage/repo.js
+++ b/src/core/storage/repo.js
@@ -1,12 +1,6 @@
 // @flow
 
-import {
-  type Disklet,
-  downgradeDisklet,
-  mapAllFiles,
-  mergeDisklets,
-  navigateDisklet
-} from 'disklet'
+import { type Disklet, mergeDisklets, navigateDisklet } from 'disklet'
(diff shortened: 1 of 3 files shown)


src/core/storage/repo.js
```diff
@@ -88,49 +84,83 @@ export async function saveChanges(
...
+  return await paths.baseDisklet
+    .setText('status.json', JSON.stringify(status))
+    .then(() => ({ status, changes }))
+}
+
+/**
+ * Lists all files in a disklet, recursively up to a limit.
+ * Returns a list of full paths.
+ */
+async function deepList(
```
swansontec, line 147: Disklet already has a helper function called `deepList`: https://github.com/EdgeApp/disklet#helpers. So, maybe we should give this one another name, like `deepListFiles`?

swansontec (approved):
Approved, but with one tiny change.

src/core/storage/repo.js
samholmes: Went with `deepListWithLimit`.

PR #395, opened 2021-09-20
Use env for webpack mode flag

samholmes:


webpack.config.js
```diff
@@ -2,8 +2,9 @@ const path = require('path')
 const webpack = require('webpack')
 const TerserPlugin = require('terser-webpack-plugin')
 
-// Set this to false for easier debugging:
-const production = true
+// DO NOT EDIT THIS CONSTANT
```
swansontec, line 5: Let's tell the use what *to* do, instead of what *not* to do. Maybe just replace these two lines with a simple `// Use "yarn prepare.dev" to make a debug-friendly build:`

package.json
```diff
@@ -62,6 +62,7 @@
     "lint": "eslint .",
     "precommit": "lint-staged && flow && npm test && npm run prepare",
     "prepare": "husky install && npm-run-all clean -p build.*",
+    "prepare.dev": "husky install && EDGE_MODE=development npm-run-all clean -p build.*",
```
swansontec, line 65: Please remove the husky bit, since husky would have already been installed by now.

swansontec (changes requested):

swansontec (approved):

PR #396, opened 2021-10-07
William/cleanups

swansontec:
I ran into a bunch of issues while trying to adjust the server Flow types. Turns out, we had a bunch of type-safety holes.

There are still a bunch of holes left to fix. but this removes a few of the gaps.


src/core/account/lobby-api.js
```diff
@@ -73,7 +73,7 @@ async function approveLoginRequest(
   }
 
   // Create a sanitized login stash object:
-  const stashTree = getStash(ai, username)
+  const { stashTree } = getStashById(ai, loginTree.loginId)
```
samholmes, line 76: What's the motivation for using `getStashById` over `getStash` here? The change incurs a BFS where there wasn't one before. Was this a bug?

samholmes (approved):
Minor change to a comment and a comment for clarity around `getStash` vs `getStashById`.

PR #397, opened 2021-10-11
William/parse ids

swansontec:
Move more parsing into the cleaners, where it is safer and more accurate to do.


src/core/login/lobby.js
```diff
@@ -168,10 +173,10 @@ export async function sendLobbyReply(
   replyData: mixed
 ): Promise<void> {
   const { io } = ai.props
-  if (lobbyRequest.publicKey == null) {
-    throw new TypeError('The lobby data does not have a public key')
+  const { publicKey } = lobbyRequest
+
```
thehobbit85, line 177: Now that the cleaner is the one throwing, isn't going to throw a different error message?
And if so, is there anyone that depends on the exact message that might break?

swansontec: Nope. This is pretty much unreachable code, unless something has gone horribly wrong.

PR #400, opened 2021-10-18
Configurable URL for the core service

samholmes:
This is useful to override the default "file://..." URL with a
localhost URL for development purposes.

---
  - https://app.asana.com/0/0/1201210979646259

### DIFF
--- a/src/io/react-native/react-native-webview.js
+++ b/src/io/react-native/react-native-webview.js
@@ -17,6 +17,7 @@ import { type YaobCallbacks, makeYaobCallbacks } from './yaob-callbacks.js'
 
 type Props = {
   allowDebugging?: boolean,
+  coreServerUrl?: string,
   debug?: boolean,
   onError(e: any): mixed,
   onLoad(clientIo: ClientIo, root: WorkerApi): Promise<mixed>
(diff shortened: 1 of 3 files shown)


src/io/react-native/react-native-webview.js
```diff
@@ -129,20 +130,21 @@ export class EdgeCoreBridge extends React.Component<Props> {
...
     }
+    console.log(`edge core at ${uri}`)
 
     return (
       <View style={debug ? styles.debug : styles.hidden}>
         <WebView
           allowFileAccess
           onMessage={this.callbacks.handleMessage}
-          originWhitelist={['file://*']}
+          originWhitelist={['file://*', 'http://*', 'https://*']}
```
swansontec, line 147: Can we infer the orign scheme from the URL, maybe like `[url.replace(/([a-z]+):\/\/.*/, "%1://*")]`

src/react-native.js
```diff
@@ -82,6 +83,7 @@ export function MakeEdgeContext(props: {
           )
           .then(onLoad)
       }
+      coreUrl={props.coreUrl}
```
swansontec, line 86: Also add this to the fake version below as well.

swansontec:
I have rebased this PR on top of the native WebView work. This works well, and maybe we can re-open it, but for now I think we don't need it anymore.

PR #403, opened 2021-10-25
William/more types

swansontec:
These are things I grabbed out of the typescript branch.


swansontec (commented):
Verbal approval from @thehobbit85

PR #404, opened 2021-10-27
Update the Bitcoin public key format

swansontec:
Based on further discovery, it appears that this specification is incorrect.

### DIFF
--- a/docs/key-formats.md
+++ b/docs/key-formats.md
@@ -50,7 +50,7 @@ The private key format is as follows:
 ```typescript
 interface PrivateBitcoinKey extends PrivateStorageKey {
   bitcoinKey: string, // Other names are possible
-  format?: 'bip32' | 'bip44' | 'bip49' | 'bip84',
+  format?: 'bip32' | 'bip44' | 'bip49',
   coinType?: number
 }
 ```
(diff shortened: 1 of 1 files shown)


samholmes:
Sending back for changes needed based on discussion in the key formats meeting today. In summary:

- Remove 'bip84' from private key formats union (only private keys with an bip49 format will be supported)
- Add specifications to the private keys section about deriving from formats:
  - Only derive from the respective format branch if bip32 or bip44
  - For bip49 we will derive from bip49 and bip84 branches
- Change the publicKey specification to reflect the union of possible values: `{ bip32: XPub } | { bip44: XPub } | { bip49: XPub, bip84: XPub }` where `XPub` is an alias for `string`.

samholmes:
Can we change the name for the `publicKey` field to at least `publicKeys` (make it plural).

I feel like the word "pub" should be used here instead, but not sure if that's appropriate (e.g. `pubs: { bip32: "xpub...." }`). The field name `xpubs` may be nice, however in the new edge-currency-plugins we are using ypubs and zpubs for the respective formats.

docs/key-formats.md
```diff
@@ -118,20 +119,48 @@ This currently happens in the core, but should move to the Bitcoin plugin.
...
+- bip49
+  - Receiving branch: xpub/0/n
+  - Change branch: xpub/1/n
+  - Script: segwit p2wpkh embedded in p2sh
+- bip84:
+  - Receiving branch: xpub/0/n
+  - Change branch: xpub/1/n
+  - Script: segwit p2wpkh
+
+It is customary for different branches to use different encoding prefixes (xpub vs ypub vs zpub), but this is not mandatory. Wallets should interpret each public key according to the table above, regardless of the encoding metadata used.
```
samholmes, line 153: As a "noob", I think being clear about the use of the word "xpub" is useful. Because xpub is interchangeably used as ether the "extended public key" no matter the serialization format and the serialization format, perhaps we should clarify that what we mean here is an extended public key either by use of another name in the derivation paths (perhaps `eXpub/0/n`, `xpubkey/0/n`, `pubkey/0/n`, `key/0/n`, or `k/0/n`). We can then mention that what this identifier means: an extended public key of any of the serialization formats (xpub, ypub, zpub, etc).

samholmes: > each public key according to the table above,

nit:

> each extended public key according to the table above,

samholmes (changes requested):
Only asking for changes with tentative feedback, or to open discussion around a comment. Otherwise, it all seems clear to me.

samholmes (approved):
/autosquash

PR #407, opened 2021-12-03
Add new types to support memo validation

Jon-edge:


src/types/types.js
```diff
@@ -378,6 +379,11 @@ export type EdgeSpendInfo = {
   otherParams?: JsonObject
 }
 
+export type MemoValidationResult =
+  | 'valid'
+  | 'invalidCharacter'
+  | 'maxLengthExceeded'
+
```
swansontec, line 386: I thought we were going to come up with an Error subtype and throw that if something went wrong, since that way we could encode more information in the future. For instance, what if we have both an invalid character and it's too long?

On the other hand, I do appreciate the simplicity of this approach. So, it's more of a question for Eliran.

PR #408, opened 2021-12-13
Meeting notes

swansontec:


src/types/types.js
```diff
@@ -682,18 +693,33 @@ export type EdgeSwapInfo = {
...
+
 export type EdgeSwapRequest = {
   // Where?
   fromWallet: EdgeCurrencyWallet,
   toWallet: EdgeCurrencyWallet,
 
   // What?
-  fromCurrencyCode: string,
-  toCurrencyCode: string,
+  fromTokenId: EdgeTokenId,
```
swansontec, line 713: fromTokenId: string,
toTokenId: string,

swansontec: TODO: Give swap plugins access to EdgeCurrencyConfig objects.

PR #412, opened 2021-12-22
William/custom webview

swansontec:
This is still WIP, but we can start testing it already. The first commit, `Replace react-native-webview with native code`, is good to go.

---
  - https://app.asana.com/0/0/1200343684549682

### DIFF
--- a/ios/EdgeCoreWebView.swift
+++ b/ios/EdgeCoreWebView.swift
@@ -147,6 +147,27 @@ class EdgeCoreWebView: RCTView, WKNavigationDelegate, WKScriptMessageHandler {
       }
       return promise.reject("Could not obtain secure entropy")
     }
+    if name == "scrypt",
+      let data64 = args[0] as? String,
+      let salt64 = args[1] as? String,
+      let n = args[2] as? UInt64,
+      let r = args[3] as? UInt32,
+      let p = args[4] as? UInt32,
+      let dklen = args[5] as? Int,
+      let data = NSData.init(base64Encoded: data64),
+      let salt = NSData.init(base64Encoded: salt64),
+      let out = NSMutableData(length: dklen)
+    {
+      if crypto_scrypt(
+        data.bytes.bindMemory(to: UInt8.self, capacity: data.length), data.length,
+        salt.bytes.bindMemory(to: UInt8.self, capacity: salt.length), salt.length,
+        n, r, p,
+        out.mutableBytes.bindMemory(to: UInt8.self, capacity: dklen), dklen
+      ) != 0 {
+        return promise.reject("Failed scrypt")
+      }
+      return promise.resolve(out.base64EncodedString())
+    }
     return promise.reject("No method \(name)")
   }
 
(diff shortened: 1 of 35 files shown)


ios/EdgeCoreWebViewManager.swift
```diff
@@ -0,0 +1,18 @@
+@objc(EdgeCoreWebViewManager) class EdgeCoreWebViewManager: RCTViewManager {
+  override static func requiresMainQueueSetup() -> Bool { return true }
+  override func view() -> UIView! { return EdgeCoreWebView() }
+  override func constantsToExport() -> [AnyHashable: Any]? {
```
swansontec, line 4: delete constants

yarn.lock
```diff
@@ -1070,6 +1077,11 @@
   resolved "https://registry.yarnpkg.com/@types/json-schema/-/json-schema-7.0.6.tgz#f4c7ec43e81b319a9815115031709f26987891f0"
   integrity sha512-3c+yGKvVP5Y9TYBEibGNR+kLtijnj7mYrXRg+WpFb2X9xm04g/DXYkfg4hmzJQosc9snFNUPkbYIhu+KAm6jJw==
 
+"@types/json-schema@^7.0.8", "@types/json-schema@^7.0.9":
```
swansontec, line 1080: TODO: Move webpack upgrade into its own commit.

ios/EdgeCoreWebView.swift
```diff
@@ -0,0 +1,232 @@
...
+      try disklet.setData(path: path, data: data)
+      return promise.resolve(nil)
+    }
+    if name == "diskletSetText",
+      let path = args[0] as? String,
+      let text = args[1] as? String
+    {
+      try disklet.setText(path: path, text: text)
+      return promise.resolve(nil)
+    }
```
paullinator, line 139: All of the Disklet methods appear to run on the main thread and would block UI rendering. In contrast, scrypt is clearly run on a separate thread/queue. This is clearly shown when debugging the app and hitting a breakpoint in Disklet code which is on Thread 1. Fix to run on separate thread/queue.

Screen Shot 2022-01-09 at 08 28 32
.

paullinator (changes requested):

ios/EdgeCoreWebView.swift
swansontec: These methods are fast, though. How much overhead is there in spinning up a thread, compared to just reading the data? I guess GCD makes the threading cheap & easy enough to be worthwhile, but I don't know if it makes as much sense on Android.

paullinator: The routines must be taking up enough of the main thread CPU time because I looked into it due to noticeable stuttering when comparing paneer versus develop . It was very slight but I am also on pretty fast phones. I imagine it would be significantly worse on older phones. 

I did only test on iOS and android might be a different story although I think it would be worth also putting that on another thread.

swansontec: Ok, I have made the change as a fixup commit, and have verified that the disklet code is running on another thread using the Xcode debugger.

For Android, the Disklet code already runs on a thread controlled by the WebView, so we don't need to do anything special there - it should not affect the GUI.

paullinator (approved):
Approved on this one commit. I'll leave it to @thehobbit85 to approve the rest of the PR

PR #423, opened 2022-01-20
Staking status

swansontec:
This is all implemented, including tests.

### DIFF
--- a/src/core/currency/wallet/currency-wallet-reducer.js
+++ b/src/core/currency/wallet/currency-wallet-reducer.js
@@ -59,6 +59,7 @@ export type CurrencyWalletState = {
 
   +paused: boolean,
 
+  +balances: EdgeBalances,
   +currencyInfo: EdgeCurrencyInfo,
   +displayPrivateSeed: string | null,
   +displayPublicSeed: string | null,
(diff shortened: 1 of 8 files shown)


test/fake/fake-currency-plugin.js
```diff
@@ -56,12 +59,22 @@ const nop: Function = () => {}
...
 }
 
+const asState = asObject({
+  balance: asOptional(asNumber),
+  stakedBalance: asOptional(asNumber),
+  tokenBalance: asOptional(asNumber),
+  blockHeight: asOptional(asNumber),
+  progress: asOptional(asNumber),
+  txs: asOptional(asMap((raw: any) => raw))
+})
```
samholmes, line 76: Shouldn't this be `asMaybe`? You were checking `typeof x === 'number'` before, now you're checking if it's not null/undefined. Not sure if it matters here in the context of the mocks, but if `changeUserSettings` receives an argument where one of the properties is defined but doesn't pass the cleaner, then there will be an exception thrown by the cleaner. Using `asMaybe` would avoid this, but this may not need to be avoided if we know that this case doesn't exist.

swansontec: That was the intent. The `changeUserSettings` method is being called make the test plugin do things. If somebody passes the wrong data type, the plugin should complain loudly, rather than silently "do nothing", which could really mess with the test.

src/core/currency/wallet/currency-wallet-reducer.js
```diff
@@ -278,8 +278,9 @@ const currencyWalletInner: FatReducer<
     state: EdgeStakingStatus = { stakedAmounts: [] },
     action: RootAction
   ): EdgeStakingStatus {
+    // Copy the object to guard against engines that mutate their data:
     return action.type === 'CURRENCY_ENGINE_CHANGED_STAKING'
-      ? action.payload.stakingStatus
+      ? { ...action.payload.stakingStatus }
```
samholmes, line 283: We ended up patching this in the plugin, but we should include this in the core to future proof this behavior. 👍

swansontec: Actually, this will only work for the top-level object. If you mutate the array inside, then we would have to splat that too...

Maybe this could be `JSON.parse(JSON.stringify())`, which would kill the problem dead. Except, it won't fix date objects, so... maybe it's better if we just let the plugins do the right thing.

samholmes (approved):

PR #428, opened 2022-01-26
William/clean account files

swansontec:
This PR applies cleaners while loading & saving files in the account data store.

---
  - https://app.asana.com/0/0/1201718130263155

### DIFF
--- a/src/core/account/account-files.js
+++ b/src/core/account/account-files.js
@@ -1,43 +1,37 @@
 // @flow
 
-import { asBoolean, asMaybe, asNumber, asObject, asOptional } from 'cleaners'
-import { type DiskletFile, type DiskletFolder, mapFiles } from 'disklet'
+import { type Disklet, justFiles } from 'disklet'
 import { base64 } from 'rfc4648'
 
-import { asBase16 } from '../../types/server-cleaners.js'
 import {
-  type EdgePluginMap,
   type EdgeWalletInfo,
   type EdgeWalletStates,
   type JsonObject
 } from '../../types/types.js'
+import { makeJsonFile } from '../../util/file-helpers.js'
 import { makeKeyInfo } from '../login/keys.js'
 import { type ApiInput } from '../root-pixie.js'
 import {
-  getStorageWalletFolder,
+  getStorageWalletDisklet,
   hashStorageWalletFilename
 } from '../storage/storage-selectors.js'
-import { type SwapSettings } from './account-reducer.js'
+import {
+  type SwapSettings,
+  asLegacyWalletFile,
+  asPluginSettingsFile,
+  asWalletStateFile
+} from './account-cleaners.js'
+
+const legacyWalletFile = makeJsonFile(asLegacyWalletFile)
+const walletStateFile = makeJsonFile(asWalletStateFile)
+const pluginSettingsFile = makeJsonFile(asPluginSettingsFile)
 
 const PLUGIN_SETTINGS_FILE = 'PluginSettings.json'
-
-type PluginSettingsFile = {
-  userSettings?: EdgePluginMap<JsonObject>,
-  swapSettings?: EdgePluginMap<SwapSettings>
-}
 
 type LoadedWalletList = {
   walletInfos: EdgeWalletInfo[],
   walletStates: EdgeWalletStates
 }
-
-const asLegacyWalletFile = asObject({
-  SortIndex: asOptional(asNumber, 0),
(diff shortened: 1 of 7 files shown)


src/core/account/data-store-api.js
```diff
@@ -1,98 +1,88 @@
...
+      const storeIds: string[] = []
+      for (const path of justFolders(await disklet.list('Plugins'))) {
+        const clean = await getFileJson(
+          disklet,
+          `${path}/Name.json`,
+          asStoreIdFile
+        )
+        if (clean != null) storeIds.push(clean.name)
+      }
+      return storeIds
```
thehobbit85, line 75: Can't we use Promise.all here?

src/util/file-helpers.js
```diff
@@ -0,0 +1,18 @@
+// @flow
+
+import { type Cleaner } from 'cleaners'
+import { type Disklet } from 'disklet'
+
+/**
+ * Reads, parses, and cleans a JSON file.
+ * Returns undefined if something goes wrong.
+ */
+export async function getFileJson<T>(
```
thehobbit85, line 12: Not that it matters too much, but I think this can "nicer" if it would be written as a "curried" function:

```js
getFileJson<T>(cleaner) => (disklet, path: string) =>  Promise<T | void>
```

That way, you could "wrap" the different cleaners into "explicit" getters like "getStoreItemFile", "getStoreIdFile" etc.

It just seems that you always use the same cleaners from the top scope and they are never really dynamic, so why not incapsulate them. But again, it doesn't really matter :)

src/core/account/account-files.js
```diff
@@ -51,43 +44,14 @@ function different(a: any, b: any): boolean {
...
-): Promise<LoadedWalletList> {
-  const files = await getJsonFiles(folder.folder('Wallets'))
-
+async function loadWalletList(disklet: Disklet): Promise<LoadedWalletList> {
   const walletInfos: EdgeWalletInfo[] = []
   const walletStates = {}
-  for (const file of files) {
-    const clean = asMaybe(asLegacyWalletFile)(file.json)
+  for (const path of justFiles(await disklet.list('Wallets'))) {
+    const clean = await getFileJson(disklet, path, asLegacyWalletFile)
```
thehobbit85, line 54: Not that this one matters too much, but can't we use Promise.all here?

thehobbit85 (changes requested):

src/core/account/data-store-api.js
```diff
@@ -1,98 +1,88 @@
...
-          .getText()
-          .then(text => JSON.parse(text).key)
-          .catch(e => undefined)
-      )
-      return itemIds.filter(itemId => typeof itemId === 'string')
+      const itemIds: string[] = []
+      for (const path of justFiles(await disklet.list(getStorePath(storeId)))) {
+        const clean = await getFileJson(disklet, path, asStoreItemFile)
+        if (clean != null) itemIds.push(clean.key)
+      }
```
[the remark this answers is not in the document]
swansontec: Ok.

src/core/account/data-store-api.js
swansontec: Ok.

src/core/account/account-files.js
swansontec: Ok.

src/core/account/account-files.js
```diff
@@ -113,17 +77,15 @@ async function loadWalletList(
...
-  })
+async function loadWalletStates(disklet: Disklet): Promise<EdgeWalletStates> {
+  const out: EdgeWalletStates = {}
+  for (const path of justFiles(await disklet.list('Keys'))) {
+    const clean = await getFileJson(disklet, path, asWalletStateFile)
+    if (clean == null) continue
+    const { id, archived, deleted, hidden, sortIndex } = clean
+    out[id] = { archived, deleted, hidden, sortIndex }
+  }
+  return out
```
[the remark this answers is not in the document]
swansontec: Ok.

src/util/file-helpers.js
swansontec: Ok, let's do it this way:

```js
const storeItemFile = makeJsonFile(asStoreItemFile)

await storeItemFile.load(disklet, path);
await storeItemFile.save(disklet, path, { ...contents })
```

This also addresses the saving side of things, which never really felt clean in the old way of doing it.

thehobbit85 (approved):
/rebase-autosquash

PR #430, opened 2022-02-01
Pass `keyOptions` to plugin `importPrivateKey` also

samholmes:


src/types/types.js
```diff
@@ -878,7 +878,7 @@ export type EdgeCurrencyConfig = {
   +userSettings: JsonObject | void,
 
   +changeUserSettings: (settings: JsonObject) => Promise<void>,
-  +importKey: (userInput: string) => Promise<JsonObject>
+  +importKey: (userInput: string, opts?: JsonObject) => Promise<JsonObject>
```
swansontec, line 881: We should not take the user-controlled options directly, but should instead take them as a named parameter: 

```
+importKey: (userInput: string, opts?: { keyOptions?: JsonObject }) => Promise<JsonObject>
```

And then you would do `const { keyOptions } = opts` later on when you call the engine.

swansontec (changes requested):

src/types/types.js
swansontec, line 881: We should not take the user-controlled JSON directly, but should instead take it as a named parameter:

```
+importKey: (userInput: string, opts?: { keyOptions?: JsonObject }) => Promise<JsonObject>
```

And then you would do `const { keyOptions } = opts` later on when you call the engine.

swansontec (changes requested):

src/core/account/plugin-api.js
```diff
@@ -60,13 +60,16 @@ export class CurrencyConfig extends Bridgeable<EdgeCurrencyConfig> {
     )
   }
 
-  async importKey(userInput: string): Promise<JsonObject> {
+  async importKey(
+    userInput: string,
+    opts?: { keyOptions: JsonObject }
```
swansontec, line 65: We prefer the `opts: {...}  = {}` where possible, rather than the `?.`. Generally, `??` and `?.` should be avoided in favor of default parameters / default destructuring where possible, since the latter are more explicit, and reduce the number of `undefined` values floating around the program as soon as possible.

swansontec (changes requested):

swansontec (approved):

PR #434, opened 2022-02-11
Add `imported` field to private keys

samholmes:
---
  - https://app.asana.com/0/0/1201810828119037

### DIFF
--- a/docs/key-formats.md
+++ b/docs/key-formats.md
@@ -38,10 +38,13 @@ All private key formats include the following two properties:
 interface PrivateStorageKey {
   dataKey: string // A 256-bit base64 encoded integer
   syncKey: string // A 160-bit base64 encoded integer
+  imported?: boolean
 }
 ```
 
 The `syncKey` uniquely identifies the wallet's Git repository on the Edge server, and the `dataKey` is the AES-256 encryption key for the data in that repository.
+
+The `imported` field should be `true` for keys that the user imported, and `false` for keys that the app generated randomly. The `imported` field is optional, and if it is missing, the keys were created before the app began recording this information.
 
 ## wallet:bitcoin
 
--- a/docs/key-formats.md
+++ b/docs/key-formats.md
@@ -1,6 +1,6 @@
 # Edge key formats
 
-The Edge account system makes it possible to backup and restore crypto-currency keys. These keys are *immutable* for safety, and there is no way to upgrade legacy keys. This means Edge support every key format customers have ever created and stored in our system, both now and for the rest of time. If changing a key format ever becomes necessary, the new format must coexist with the old format, and the wallet must be able to read both.
+The Edge account system makes it possible to backup and restore crypto-currency keys. These keys are _immutable_ for safety, and there is no way to upgrade legacy keys. This means Edge support every key format customers have ever created and stored in our system, both now and for the rest of time. If changing a key format ever becomes necessary, the new format must coexist with the old format, and the wallet must be able to read both.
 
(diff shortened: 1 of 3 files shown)


docs/key-formats.md
```diff
@@ -83,6 +84,8 @@ If `format` is missing, the wallet uses a default of bip32.
 
 If the `coinType` is missing, the default is 0 for Bitcoin. Other coins use different defaults for `coinType`.
 
+The `imported` field signifies whether the wallet was imported or created. If the `imported` field is `undefined`, then the wallet must have been created before this field was added to the specification and whether the wallet was imported or created is unknown. Implementations should specify `true` if the wallet was imported, and `false` if the wallet was created moving forward.
+
```
swansontec, line 88: I would word this differently:

> The `imported` field should be `true` for keys that the user imported, and `false` for keys that the app generated randomly. The `imported` field is optional, and if it is missing, the keys were created before the app began recording this information.

Also, I would move this paragraph to the "Storage keys" section, which should then be renamed to "Common key properties" or such.

src/core/account/plugin-api.js
```diff
@@ -69,7 +69,9 @@ export class CurrencyConfig extends Bridgeable<EdgeCurrencyConfig> {
     if (tools.importPrivateKey == null) {
       throw new Error('This wallet does not support importing keys')
     }
-    return tools.importPrivateKey(userInput, opts.keyOptions)
+    const keys = await tools.importPrivateKey(userInput, opts.keyOptions)
+    keys.imported = true
+    return keys
```
swansontec, line 74: Can we change this structure to `return { ...keys, imported: true }`. That will keep TypeScript happy after we run the conversion script.

src/core/login/keys.js
```diff
@@ -319,8 +319,10 @@ export async function createCurrencyWallet(
       throw new Error('This wallet does not support importing keys')
     }
     keys = await tools.importPrivateKey(opts.importText, opts.keyOptions)
+    keys.imported = true
```
swansontec, line 322: `keys = { ...await tools.importPrivateKey(), imported: true }`

src/core/login/keys.js
```diff
@@ -319,8 +319,10 @@ export async function createCurrencyWallet(
...
   } else {
     keys = await tools.createPrivateKey(walletType, opts.keyOptions)
+    keys.imported = false
```
swansontec, line 325: `keys = { ...await tools.createPrivateKey(), imported: false }`

swansontec (changes requested):

swansontec (approved):

src/core/login/keys.js
```diff
@@ -318,9 +318,15 @@ export async function createCurrencyWallet(
...
-    keys = await tools.importPrivateKey(opts.importText, opts.keyOptions)
+    keys = {
+      ...(await tools.importPrivateKey(opts.importText, opts.keyOptions)),
+      imported: true
+    }
   } else {
-    keys = await tools.createPrivateKey(walletType, opts.keyOptions)
+    keys = {
+      ...(await tools.createPrivateKey(walletType, opts.keyOptions)),
+      imported: true
```
swansontec, line 328: This one needs to be false.

swansontec (changes requested):
You got the flag backwards in your fixup.

src/core/login/keys.js
samholmes: Oops. That was a derp on my part. Fixed up.

swansontec (approved):

PR #435, opened 2022-02-14
William/edge token

swansontec:
This adds the new `EdgeToken` type to the core. The token management logic still needs to be copied over from the GUI.


src/core/account/account-reducer.js
```diff
@@ -249,6 +251,18 @@ const accountInner: FatReducer<
     return action.type === 'LOGIN' ? action.payload.username : state
   },
 
+  customTokens(state = {}, action: RootAction): EdgePluginMap<SwapSettings> {
```
swansontec, line 254: `SwapSettings` should be `EdgeTokenMap`, which means Flow is pretty broken here.

swansontec: Found a way to fix it, updated the PR.

PR #440, opened 2022-02-22
Bug fixes

swansontec:
Here is my version of #439.

### DIFF
--- a/src/core/account/account-pixie.js
+++ b/src/core/account/account-pixie.js
@@ -223,17 +223,17 @@ const accountPixie: TamePixie<AccountProps> = combinePixies({
       let dirty = lastActiveWalletIds !== activeWalletIds
       lastActiveWalletIds = activeWalletIds
 
-      let lastOut = {}
+      let lastOut: { [walletId: string]: EdgeCurrencyWallet } = {}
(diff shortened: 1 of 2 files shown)


src/core/account/account-pixie.js
```diff
@@ -223,17 +223,17 @@ const accountPixie: TamePixie<AccountProps> = combinePixies({
...
 
-      const out = {}
+      const out: { [walletId: string]: EdgeCurrencyWallet } = {}
       const { wallets } = input.props.output.currency
       for (const walletId of activeWalletIds) {
-        const api = wallets?.[walletId]?.walletApi
+        const api = wallets[walletId]?.walletApi
         if (api !== lastOut[walletId]) dirty = true
-        out[walletId] = api
+        if (api != null) out[walletId] = api
```
thehobbit85, line 236: Wouldn't it be safer if you added this after 234?
`if (api == null) continue`

thehobbit85 (changes requested):

src/core/account/account-pixie.js
swansontec: No, that would be incorrect. We want to do the diff in every case, in case the wallet gets archived & then restored.

thehobbit85 (approved):

PR #442, opened 2022-03-01
William/custom tokens

swansontec:
This moves the loading & saving of custom tokens into the core, instead of the GUI.

This is not a breaking change, since the core will just ignore already-existing tokens if the GUI adds them. Once we publish this, we can simply delete a bunch of code from the GUI.

---
  - https://app.asana.com/0/0/1201833665439726

### DIFF
--- a/src/core/account/account-cleaners.js
+++ b/src/core/account/account-cleaners.js
@@ -2,6 +2,7 @@
 
 import {
   type Cleaner,
+  asArray,
   asBoolean,
   asMap,
   asNumber,
@@ -11,12 +12,26 @@ import {
 } from 'cleaners'
 
 import { asBase16 } from '../../types/server-cleaners.js'
+import { type EdgeDenomination, type EdgeToken } from '../../types/types.js'
 import { asJsonObject } from '../../util/file-helpers.js'
 import { type SwapSettings } from './account-types.js'
 
 // ---------------------------------------------------------------------
 // building-block types
 // ---------------------------------------------------------------------
+
+const asEdgeDenomination: Cleaner<EdgeDenomination> = asObject({
+  multiplier: asString,
+  name: asString,
+  symbol: asOptional(asString)
+})
+
+const asEdgeToken: Cleaner<EdgeToken> = asObject({
+  currencyCode: asString,
+  denominations: asArray(asEdgeDenomination),
+  displayName: asString,
+  networkLocation: asOptional(asJsonObject)
+})
 
 const asSwapSettings: Cleaner<SwapSettings> = asObject({
   enabled: asOptional(asBoolean, true)
@@ -58,3 +73,25 @@ export const asPluginSettingsFile = asObject({
   // Swap plugins:
   swapSettings: asOptional(asMap(asSwapSettings), {})
 }).withRest
+
+/**
+ * The settings file managed by the GUI.
+ */
+export const asGuiSettingsFile = asObject({
+  customTokens: asArray(
+    asObject({
+      contractAddress: asString,
+      currencyCode: asString,
+      currencyName: asString,
+      denomination: asString,
+      denominations: asArray(asEdgeDenomination),
+      isVisible: asOptional(asBoolean, true),
+      multiplier: asString,
+      walletType: asOptional(asString, 'wallet:ethereum')
+    })
+  )
+})
+
+export const asCustomTokensFile = asObject({
+  customTokens: asObject(asObject(asEdgeToken))
+})
--- a/src/core/account/account-reducer.js
+++ b/src/core/account/account-reducer.js
@@ -266,6 +266,10 @@ const accountInner: FatReducer<
 
   customTokens(state = {}, action: RootAction): EdgePluginMap<EdgeTokenMap> {
     switch (action.type) {
+      case 'ACCOUNT_CUSTOM_TOKENS_LOADED': {
+        const { customTokens } = action.payload
+        return customTokens
+      }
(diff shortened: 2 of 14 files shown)


src/core/account/custom-tokens.js
```diff
@@ -91,3 +145,16 @@ export async function loadBuiltinTokens(
...
   )
 }
+
+function findEngine(ai: ApiInput, pluginId: string): EdgeCurrencyEngine | void {
+  for (const walletId of Object.keys(ai.props.state.currency.wallets)) {
+    const walletOutput = ai.props.output.currency.wallets[walletId]
+    if (
+      walletOutput != null &&
+      walletOutput.engine != null &&
+      ai.props.state.currency.wallets[walletId].pluginId === pluginId
```
samholmes, line 167: If you use `const [walletId, wallet] = Object.entries(...)` in the loop you can simplify this condition to:

```js
      wallet.pluginId === pluginId
```

src/core/account/plugin-api.js
```diff
@@ -59,6 +61,41 @@ export class CurrencyConfig extends Bridgeable<EdgeCurrencyConfig> {
...
+    })
+    return tokenId
+  }
+
+  async changeCustomToken(tokenId: string, token: EdgeToken): Promise<void> {
+    const { _accountId: accountId, _ai: ai, _pluginId: pluginId } = this
+    const newTokenId = await getTokenId(ai, pluginId, token)
+    if (newTokenId !== tokenId) {
+      throw new Error(
+        `The tokenId would change from ${tokenId} to ${newTokenId}`
```
samholmes, line 80: Maybe:

```js
throw new Error(`Mismatching tokenId '${newTokenId}' when updating for tokenId '${tokenId}'`)
```

or similar

src/core/account/account-cleaners.js
```diff
@@ -58,3 +73,25 @@ export const asPluginSettingsFile = asObject({
...
+      denominations: asArray(asEdgeDenomination),
+      isVisible: asOptional(asBoolean, true),
+      multiplier: asString,
+      walletType: asOptional(asString, 'wallet:ethereum')
+    })
+  )
+})
+
+export const asCustomTokensFile = asObject({
+  customTokens: asMap(asMap(asEdgeToken))
```
samholmes, line 96: Hasn't `asMap` been replaced in favor of `asObject`?

samholmes (approved):

samholmes:
I meant to submit feedback as comments, but instead click approve. However, the feedback is not major, so I would approve if all the feedback is rejected.

src/core/account/account-cleaners.js
swansontec: Derp. I'll fix this.

src/core/account/plugin-api.js
swansontec: I think the existing message is better, because something has changed (like the contract address) that affects the token's identity. Otherwise, we could say that this is actually OK, drop the error message, and just treat this as an add / remove operation.

src/core/account/custom-tokens.js
swansontec: I don't think we we can use `Object.entries` in the core. Old Androids don't have it, and we don't have a polyfill.

src/core/account/account-reducer.js
```diff
@@ -267,10 +268,24 @@ const accountInner: FatReducer<
...
         return { ...state, [pluginId]: newList }
       }
+      case 'ACCOUNT_CUSTOM_TOKEN_REMOVED': {
+        const { pluginId, tokenId } = action.payload
+        const oldList = state[pluginId] ?? {}
+
+        // Has anything changed?
+        if (oldList[tokenId] == null) return state
+
+        const { [tokenId]: unused, ...newList } = oldList
```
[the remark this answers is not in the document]
swansontec: Right there. It's a destructuring assignment, except that we don't really care what the value is, so I assign it to a variable named `unused` which really is unused anywhere else.

src/core/root-pixie.js
```diff
@@ -37,6 +37,8 @@ export type RootProps = {
 
 export type ApiInput = PixieInput<RootProps>
 
+export const toApiInput = (input: PixieInput<any>): ApiInput => input
```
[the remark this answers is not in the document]
swansontec: I will add it.

peachbits (approved):

src/core/account/custom-tokens.js
```diff
@@ -20,6 +22,45 @@ const asMaybeContractLocation = asMaybe(
   })
 )
 
+/**
+ * We need to validate the token before we can add it.
+ *
+ * If the plugin has a `getTokenId` method, just use that.
+ *
+ * Otherwise, we need to call `EdgeCurrencyEngine.addCustomToken`
+ * to validate the contract address, and then guess the tokenId from that.
```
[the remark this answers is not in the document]
swansontec: Indeed, the only things we validate right now is the contract address is hexadecimal, and that the multiplier is a number. Even so, those are valuable. Going forward, we can do things like what you are saying for the contract name & other info.

I wonder if we need another method on `EdgeCurrencyTools` to look up extra information, like the multiplier. Something take `autoFillToken: (userInput: Partial<EdgeToken>) => Promise<EdgeToken>`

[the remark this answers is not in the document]
thehobbit85: I agree that this is useful but there is already a separate task for this and it should not be part of this PR.

PR #447, opened 2022-03-30
William/fixes

swansontec:
Here are various core issues I ran into while doing my super-deep dive into the performance.


src/core/currency/wallet/currency-wallet-files.js
```diff
@@ -341,7 +341,7 @@ async function loadAddressFiles(input: CurrencyWalletInput): Promise<void> {
   )
 
   // Load these addresses into the engine:
-  const engine = input.props.walletOutput.engine
+  const engine = input.props.walletOutput?.engine
```
thehobbit85, line 344: Great catch!

thehobbit85 (approved):

PR #455, opened 2022-05-11
Await the engine's addCustomToken method so errors can be handled

peachbits:
---
  - https://app.asana.com/0/0/1202239521389815


swansontec (commented):
This is actually the correct change. Can you reopen this so we can merge it?

swansontec (approved):
Thanks for finding this genuine "oops".

PR #459, opened 2022-06-09
Sam/confirmations

samholmes:


src/core/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -189,6 +189,26 @@ export function makeCurrencyWalletCallbacks(
...
+          // Default confirmation rule is 1 block
+          const {
+            requiredConfirmations = 1
+          } = input.props.walletState.currencyInfo
+          const { height } = input.props.walletState
+          const blockConfirmations = height - tx.blockHeight
+
+          tx.confirmations =
+            height === 0
+              ? null
```
thehobbit85, line 202: Confirmations cannot be 'null'.
Look at the type and also the precommit tests fails.

thehobbit85 (changes requested):

src/core/currency/wallet/currency-wallet-callbacks.js
samholmes: Fixed. I meant to keep it `undefined` if the block height for the wallet is 0 (not synchronized).

thehobbit85 (approved):

PR #465, opened 2022-07-19
Change login server to login.edge.app

paullinator:
---
  - https://app.asana.com/0/0/1203326088773348

### DIFF
--- a/src/core/root.js
+++ b/src/core/root.js
@@ -20,6 +20,8 @@ import { type RootState, defaultLogSettings, reducer } from './root-reducer.js'
 
 let allContexts: EdgeContext[] = []
 
+const ACCEPTED_SERVER_DOMAINS = ['edge.app', 'edgetest.app', 'localhost']
+
 const composeEnhancers =
   typeof window === 'object' && window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__
     ? window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__({ name: 'core' })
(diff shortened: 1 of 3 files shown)


swansontec (approved):

src/core/root.js
```diff
@@ -39,7 +39,7 @@ export async function makeContext(
   const {
     apiKey,
     appId = '',
-    authServer = 'https://auth.airbitz.co/api',
+    authServer = 'https://login.edge.app/api',
```
swansontec, line 42: You missed a spot: src/react-native.js, line 148.

swansontec (changes requested):

paullinator:
I decided to expand the scope of this PR to also restrict login server overrides to *.edge.app, *.edgetest.app and localhost domains

src/core/root.js
```diff
@@ -20,6 +20,8 @@ import { type RootState, defaultLogSettings, reducer } from './root-reducer.js'
 
 let allContexts: EdgeContext[] = []
 
+const ACCEPTED_SERVER_DOMAINS = ['edge.app', 'edgetest.app', 'localhost']
```
swansontec, line 23: This is not the approach we agreed on back in the day - we were going to have the server list hard-coded internally, and then accept an enum like `'localhost' | 'test' | 'production'` to select among the limited set of options. This would eventually allow us to hook into the info server as well, since we have a set of known keys to trigger off of.

In either case, this all seems out-of-scope for this PR, unless you want to prioritize this right away as a P1.

swansontec (changes requested):

src/core/root.js
paullinator: I do like the enum for the sake of triggering off the info server. But for development/testing, the open string seems more flexible.

How about `authServer === undefined` implies we can grab off the info server. But if you pass a string, we validate and use that? Seems to cover all use cases.

swansontec (approved):

PR #467, opened 2022-07-25
Allow plugins to fail to load

peachbits:
This was tested by breaking a currency plugin on devices that don't support bigints. It does not blow up. The app loads normally without any mention of the broken plugin.

This has not been tested with other types of plugins like rates or swaps.

---
  - https://app.asana.com/0/0/1202654362524309

### DIFF
--- a/src/core/plugins/plugins-reducer.js
+++ b/src/core/plugins/plugins-reducer.js
@@ -44,6 +44,14 @@ export const plugins = (
       }
       for (const pluginId of Object.keys(action.payload)) {
         const plugin = action.payload[pluginId]
+
+        // Don't stop loading the bundle if plugin(s) fail to load. Some plugins may rely on advanced features
+        // that aren't locally available so we can skip loading those but should continue and load what we can.
+        if (plugin == null) {
+          out.init = { ...out.init, [pluginId]: false }
+          continue
+        }
+
         // $FlowFixMe - Flow doesn't see the type refinement here:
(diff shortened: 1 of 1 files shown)


swansontec (changes requested):
Nope! This is not the way.

You know how plugins have a factory function that takes the IO objects? We should add `| void` to the allowed return values of that function. That way, your `makePolkadotPlugin` can decide whether or not the browser is good to go, and then return `undefined` if it isn't happy. This is different from allowing plugins to crash, and then just ignoring the errors.

peachbits:
It is definitely a hack I just needed to get into QA so they test old androids. Your suggestion makes sense I'll look into it and update the PR

src/core/plugins/plugins-reducer.js
```diff
@@ -44,6 +44,14 @@ export const plugins = (
       }
       for (const pluginId of Object.keys(action.payload)) {
         const plugin = action.payload[pluginId]
+
+        // Don't stop loading the bundle if plugin(s) fail to load. Some plugins may rely on advanced features
+        // that aren't locally available so we can skip loading those but should continue and load what we can.
+        if (plugin == null) {
+          delete out.init[pluginId]
```
swansontec, line 51: This is Redux, so we can't just delete things. I think we can do `out.init = { ...out.init, [pluginId]: false }`, though.

swansontec (changes requested):

swansontec (approved):

PR #471, opened 2022-08-19
Create and save a random `clientId` on first run

swansontec:


src/core/root.js
```diff
@@ -11,6 +11,7 @@ import {
   type EdgeRateHint
 } from '../types/types.js'
 import { type RootAction } from './actions.js'
+import { CLIENT_FILE_NAME, clientFile } from './context/client-file.js'
```
samholmes, line 14: Missing file `./context/client-file.js`

samholmes (changes requested):

samholmes (approved):

PR #479, opened 2022-11-02
Accept tokenId's for spending and swapping

swansontec:
We will upgrade the `currencyCode` into the `tokenId` if not provided, and vice-versa so the plugin remains compatible as well.

---
  - https://app.asana.com/0/0/1203287918505521

### DIFF
--- a/src/types/error.js
+++ b/src/types/error.js
@@ -4,7 +4,8 @@ import { type Cleaner, asMaybe } from 'cleaners'
 import { base64 } from 'rfc4648'
 
 import { asOtpErrorPayload, asPasswordErrorPayload } from './server-cleaners.js'
-import type { EdgeSwapInfo } from './types.js'
+import { upgradeCurrencyCode } from './type-helpers.js'
+import type { EdgeSwapInfo, EdgeSwapRequest } from './types.js'
 
 /*
  * These are errors the core knows about.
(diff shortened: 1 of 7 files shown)


src/core/account/custom-tokens.js
```diff
@@ -247,3 +247,27 @@ export async function saveCustomTokens(
...
+  walletId: string,
+  info: { currencyCode?: string, tokenId?: string }
+): { currencyCode: string, tokenId?: string } {
+  const walletState = ai.props.state.currency.wallets[walletId]
+  const { currencyInfo, pluginId } = walletState
+  const allTokens = ai.props.state.accounts[accountId].allTokens[pluginId]
+
+  const tokenId =
+    info.tokenId ??
+    Object.keys(allTokens).find(
```
swansontec, line 267: Oops. This doesn’t handle when the passed currency code is for the chain itself, like “BTC”

src/types/error.js
```diff
@@ -264,19 +264,45 @@ export class SwapCurrencyError extends Error {
...
+            request.fromWallet.currencyInfo.currencyCode
+      toCurrency =
+        request.toTokenId != null
+          ? request.toWallet.currencyConfig.allTokens[request.toTokenId]
+              .currencyCode
+          : request.toCurrencyCode ?? request.toWallet.currencyInfo.currencyCode
+    }
+
     super(
       `${swapInfo.displayName} does not support ${fromCurrency} to ${toCurrency}`
```
paullinator, line 296: Should probably show the tokenIds as well to differentiate what was trying to be used. That's the purpose of the error.

src/types/error.js
```diff
@@ -264,19 +264,45 @@ export class SwapCurrencyError extends Error {
...
+    if (typeof request === 'string') {
+      fromCurrency = request
+      if (toCurrency == null) toCurrency = 'unknown'
+    } else {
+      fromCurrency =
+        request.fromTokenId != null
+          ? request.fromWallet.currencyConfig.allTokens[request.fromTokenId]
+              .currencyCode
+          : request.fromCurrencyCode ??
+            request.fromWallet.currencyInfo.currencyCode
```
paullinator, line 287: why fallback to `request.fromWallet.currencyInfo.currencyCode`? Shouldn't the request.fromCurrencyCode always exist?

paullinator (changes requested):

src/types/error.js
swansontec: The type system demands it. But I could put `''` or something if that would be simpler.

src/types/error.js
swansontec: The error is meant to be translated by the GUI into something more usable, which is why we put the tokenId's into the data part. I can adjust the message to include those as well, in case the GUI doesn't do the translation.

src/types/error.js
paullinator: I think some invalid value would be better so we don't imply that is an expected code-flow. `` would be fine.

src/types/error.js
paullinator: That would be great. Also helps for the logs that gets spewed.

swansontec:
@paullinator I have had to do some refactoring to address your issues. I moved the `upgradeCurrencyCode` into its own file, then took advantage of it inside the `SwapCurrencyError` constructor. This new file needs to be included in the Flow -> TS conversion script, since errors.js depends on it. Finally, I fixed `upgradeCurrencyCode` to address cases where the currency code matches the parent currency.

paullinator (approved):

paullinator:
looks good.

PR #483, opened 2022-11-03
Add the swap request to the quote

swansontec:
This makes the quote more self-contained.

---
  - https://app.asana.com/0/0/1203287918505533

### DIFF
--- a/src/types/types.js
+++ b/src/types/types.js
@@ -858,6 +858,8 @@ export type EdgeSwapApproveOptions = {
  * If a provider can satisfy a request, what is their price?
  */
 export type EdgeSwapQuote = {
+  +request: EdgeSwapRequest,
+
   +isEstimate: boolean,
   +fromNativeAmount: string,
   +toNativeAmount: string,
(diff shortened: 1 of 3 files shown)


src/types/types.js
```diff
@@ -858,6 +858,8 @@ export type EdgeSwapApproveOptions = {
  * If a provider can satisfy a request, what is their price?
  */
 export type EdgeSwapQuote = {
+  +request: EdgeSwapRequest,
```
paullinator, line 861: Shouldn't this be optional so that we don't break all the plugins that return an EdgeSwapQuote but without the `request` object

paullinator (changes requested):

src/types/types.js
swansontec: Right, but notice that we jam the request into the object on the way out to the GUI. So, if we don't update the plugins, the GUI still gets the object. This will be a breaking change for the swap plugin from a types perspective, but it will be an easy fix.

paullinator (approved):

PR #484, opened 2022-11-10
Replace EdgeTransaction['wallet'] with EdgeTransaction['walletId']

samholmes:
---
  - https://app.asana.com/0/0/1203342164826757

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # edge-core-js
 
+## Unreleased
+
+- added: `EdgeTransaction.walletId`.
+- deprecated: `EdgeTransaction.wallet`. Use `EdgeTransaction.walletId` instead.
+
 ## v0.19.31 (2022-11-02)
 
 - added: Specifying token spends by their ID, instead of their imprecise currency code:
--- a/src/core/currency/wallet/currency-wallet-api.js
+++ b/src/core/currency/wallet/currency-wallet-api.js
@@ -380,7 +380,10 @@ export function makeCurrencyWalletApi(
           }
           // add this tx / file to the output
           const edgeTx = combineTxWithFile(input, tx, file, currencyCode)
-          if (searchStringFilter(edgeTx, opts) && dateFilter(edgeTx, opts)) {
+          if (
+            searchStringFilter(ai, edgeTx, opts) &&
+            dateFilter(edgeTx, opts)
+          ) {
             out.push(edgeTx)
           }
           searchedTxs++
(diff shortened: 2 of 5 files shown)


src/core/currency/wallet/currency-wallet-api.js
```diff
@@ -639,6 +645,7 @@ export function combineTxWithFile(
   currencyCode: string
 ): EdgeTransaction {
   const wallet = input.props.walletOutput.walletApi
+  const walletId = wallet?.id ?? '' // Let's just assume the wallet is defined
```
swansontec, line 648: You can use `input.props.walletId` here, instead of grabbing it off the wallet.

src/types/types.js
```diff
@@ -398,7 +398,7 @@ export type EdgeTransaction = {
 
   // Core:
   metadata?: EdgeMetadata,
-  wallet?: EdgeCurrencyWallet, // eslint-disable-line no-use-before-define
```
swansontec, line 401: I think we need to add the block at the end of the interface:

```
// Deprecated:
wallet?: EdgeCurrencyWallet // We never populate this
```

That way, it's not a breaking change.

src/core/currency/wallet/currency-wallet-export.js
```diff
@@ -15,10 +16,16 @@ export function dateFilter(
   return false
 }
 
+export type SearchStringFilterContext = {
+  currencyState: CurrencyState
+}
+
 export function searchStringFilter(
+  context: SearchStringFilterContext,
```
swansontec, line 24: If a function needs access to state inside the core, we usually pass an initial `ai: ApiInput` parameter, which contains redux state + dispatch.

However, in this case, it might make more sense to just pass the wallet. The only places where we call `searchStringFilter` as wallet-specific, and the wallet should have everything we could ever need to interpret the transaction.

Either option would be fine.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # edge-core-js
 
+## Unreleased
+
+- Replace optional `wallet` field from `EdgeTransaction` with required `walletId`
```
swansontec, line 5: Going with the semantic changelog thing, it should be:

```
- added: `EdgeTransaction.walletId`.
- deprecated: `EdgeTransaction.wallet`. Use `EdgeTransaction.walletId` instead.
```

swansontec (changes requested):

src/core/currency/wallet/currency-wallet-export.js
samholmes: I went with `ai: ApiInput` only because I don't know where the wallet is stored in the data model.

swansontec (changes requested):
One leftover

swansontec (approved):

PR #486, opened 2022-11-16
Fix login server override testing

swansontec:
This was allowing `"https://fakeedge.app"` and similar domains, so now it uses a proper parser instead of doing string manipulation.

We have historically put our testing servers on *.edge.app, so I have also dropped the "edgetest.app" domain. We can add it back another day if we decide to deploy servers there.

---
  - https://app.asana.com/0/0/1203395352450560

### DIFF
--- a/src/core/root.js
+++ b/src/core/root.js
@@ -20,8 +20,6 @@ import { type RootState, defaultLogSettings, reducer } from './root-reducer.js'
 
 let allContexts: EdgeContext[] = []
 
-const ACCEPTED_SERVER_DOMAINS = ['edge.app', 'edgetest.app', 'localhost']
-
 const composeEnhancers =
   typeof window === 'object' &&
   window.__REDUX_DEVTOOLS_EXTENSION_COMPOSE__ != null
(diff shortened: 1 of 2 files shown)


src/core/root.js
```diff
@@ -155,3 +147,22 @@ export function closeEdge(): void {
...
+ */
+export function validateServer(server: string): void {
+  const url = new URL(server)
+
+  if (url.protocol === 'http:') {
+    if (url.hostname === 'localhost') return
+  }
+  if (url.protocol === 'https:') {
+    if (url.hostname === 'localhost') return
+    if (/^([A-Za-z0-9_-]+\.)*edge\.app$/.test(url.hostname)) return
```
paullinator, line 162: needs to also support edgetest.app

paullinator (changes requested):

src/core/root.js
swansontec: I put my rationale in both the commit message and the pull request. We have never used edgetest for anything, so I would like to remove it until we have a real reason to add it back. We consider our login server to be sensitive infrastructure (one of the few repos we don't make open source), so it will always be deployed `*.edge.app`, such as `login-test.edge.app`.

paullinator (approved):

PR #487, opened 2022-11-17
Use modern syntax

swansontec:
These big style changes make life better with TypeScript + ESLint.

---
  - https://app.asana.com/0/0/1203395352450567

### DIFF
--- a/src/core/account/account-files.js
+++ b/src/core/account/account-files.js
@@ -120,9 +120,10 @@ export async function loadAllWalletStates(
   )
 
   // Merge all that information together:
-  const legacyWalletInfos: EdgeWalletInfo[] = [].concat(
-    ...legacyLists.map(files => files.walletInfos)
-  )
+  const legacyWalletInfos: EdgeWalletInfo[] = []
+  for (const files of legacyLists) {
+    legacyWalletInfos.push(...files.walletInfos)
+  }
   const legacyWalletStates: EdgeWalletStates[] = legacyLists.map(
     files => files.walletStates
   )
--- a/src/core/account/account-files.js
+++ b/src/core/account/account-files.js
@@ -163,7 +163,7 @@ export async function changeWalletStates(
 
   // If there are no changes, do nothing:
   const walletIds = Object.keys(toWrite)
-  if (!walletIds.length) return
+  if (walletIds.length < 1) return
 
   await Promise.all(
     walletIds.map(async walletId => {
--- a/src/core/account/account-pixie.js
+++ b/src/core/account/account-pixie.js
@@ -131,7 +131,7 @@ const accountPixie: TamePixie<AccountProps> = combinePixies({
           accountWalletInfos.map(info => syncStorageWallet(ai, info.id))
         )
         const changes: string[] = [].concat(...changeLists)
-        if (changes.length) {
+        if (changes.length > 0) {
           await Promise.all([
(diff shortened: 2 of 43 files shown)


src/core/account/account-files.js
```diff
@@ -120,9 +120,10 @@ export async function loadAllWalletStates(
   )
 
   // Merge all that information together:
-  const legacyWalletInfos: EdgeWalletInfo[] = [].concat(
-    ...legacyLists.map(files => files.walletInfos)
-  )
+  const legacyWalletInfos: EdgeWalletInfo[] = []
+  for (const files of legacyLists) {
+    legacyWalletInfos.push(...files.walletInfos)
+  }
```
samholmes, line 126: Why not:

```ts
const legacyWalletInfos: EdgeWalletInfo[] = legacyLists.map(files => files.walletInfos)
```

src/core/account/account-pixie.js
```diff
@@ -130,7 +130,8 @@ const accountPixie: TamePixie<AccountProps> = combinePixies({
         const changeLists = await Promise.all(
           accountWalletInfos.map(info => syncStorageWallet(ai, info.id))
         )
-        const changes: string[] = [].concat(...changeLists)
+        const changes: string[] = []
```
samholmes, line 133: Also why isn't this simpler:

```ts
const changes: string[] = [...changeLists]
```

src/core/currency/wallet/currency-wallet-export.js
```diff
@@ -75,7 +75,7 @@ export function searchStringFilter(
       )
         return true
     }
-    if (tx.swapData != null && tx.swapData.plugin != null) {
+    if (tx.swapData != null) {
```
samholmes, line 78: Did you mean `tx.swapData?.plugin != null`?

samholmes (commented):
Some questionable feedback

src/core/account/account-files.js
swansontec: That is not the same result. Notice the `...` within the `push`, which has the effect of concatenating all these individual arrays together into one big array.

src/core/account/account-pixie.js
swansontec: That is not the same result. We are concatenating a bunch of little arrays into one big array.

src/core/currency/wallet/currency-wallet-export.js
swansontec: That was the original replacement, yes, but Flow blew up here for some reason (it works fine in the other places). I poked around in the code, and since `swapData` comes from a cleaner, `plugin` can never be null anyhow.

src/core/account/account-files.js
samholmes: Oh it's a `.flatMap` operation. Okay, then your solution is fine because flow doesn't think `Array.prototype.flatMap` is a thing.

src/core/currency/wallet/currency-wallet-export.js
samholmes: I see. That makes sense.

samholmes (approved):

src/core/account/account-files.js
swansontec: Also, we can't use flatMap in the core because old Androids.

PR #489, opened 2022-11-20
Paul/new address opts

paullinator:
---
  - https://app.asana.com/0/0/1203395121961357

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.js
+++ b/src/core/currency/wallet/currency-wallet-api.js
@@ -15,6 +15,7 @@ import {
   type EdgeCurrencyWallet,
   type EdgeDataDump,
   type EdgeEncodeUri,
+  type EdgeGetReceiveAddressOptions,
   type EdgeGetTransactionsOptions,
   type EdgeMemoRules,
   type EdgeMetadata,
@@ -394,15 +395,13 @@ export function makeCurrencyWalletApi(
     },
 
     async getReceiveAddress(
-      opts: EdgeCurrencyCodeOptions = {}
+      opts: EdgeGetReceiveAddressOptions = {}
     ): Promise<EdgeReceiveAddress> {
       const freshAddress = await engine.getFreshAddress(opts)
       const receiveAddress: EdgeReceiveAddress = {
-        metadata: fakeMetadata,
+        ...freshAddress,
         nativeAmount: '0',
-        publicAddress: freshAddress.publicAddress,
-        legacyAddress: freshAddress.legacyAddress,
-        segwitAddress: freshAddress.segwitAddress
+        metadata: fakeMetadata
       }
       return receiveAddress
     },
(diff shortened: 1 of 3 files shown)


src/types/types.js
```diff
@@ -456,8 +456,11 @@ export type EdgeDataDump = {
 
 export type EdgeFreshAddress = {
   publicAddress: string,
+  nativeAmount: string,
   segwitAddress?: string,
-  legacyAddress?: string
+  legacyAddress?: string,
+  segwitNativeAmount?: string,
+  legacyNativeAmount?: string
```
swansontec, line 463: We have some confusion here, since we were using `nativeAmount` before to indicate the amount requested, but now we are using it to indicate the balance. So maybe we should call these fields `nativeBalance?: string`, making them optional so we can tell the difference between 0 and "I don't know".

Otherwise, if you know for sure that we will never care about the difference between 0 and "not reported", we could make the balances mandatory, but then we would need fallback code to upgrade the responses from plugins that haven't been upgraded yet.

src/core/currency/wallet/currency-wallet-api.js
```diff
@@ -394,15 +395,12 @@ export function makeCurrencyWalletApi(
...
     ): Promise<EdgeReceiveAddress> {
       const freshAddress = await engine.getFreshAddress(opts)
       const receiveAddress: EdgeReceiveAddress = {
-        metadata: fakeMetadata,
-        nativeAmount: '0',
-        publicAddress: freshAddress.publicAddress,
-        legacyAddress: freshAddress.legacyAddress,
-        segwitAddress: freshAddress.segwitAddress
+        ...freshAddress,
+        metadata: fakeMetadata
```
swansontec, line 403: If we want the `nativeAmount` to be mandatory, existing plugins won't be sending that over, so we are going to have a problem. You can do a run-time fallback of `{ nativeAmount: '0', ...freshAddress, metadata: fakeMetadata }` to correct this.

However, as mentioned in the other comment, I think we should do `nativeBalance?: string` to bypass the whole problem.

swansontec (changes requested):
This idea seems good, but we don't provide the right fallback logic.

src/types/types.js
paullinator: Ah. I never realized that it meant amount requested! Sam pointed it out and he thought it was the address balance. I'll change to `nativeBalance` and make them all optional.

src/core/currency/wallet/currency-wallet-api.js
paullinator: I agree. Moved to nativeBalance

test/fake/fake-currency-plugin.js
```diff
@@ -262,9 +263,9 @@ class FakeCurrencyEngine {
 
   // Addresses:
   async getFreshAddress(
-    opts: EdgeCurrencyCodeOptions
+    opts: EdgeGetReceiveAddressOptions
   ): Promise<EdgeFreshAddress> {
-    return { publicAddress: 'fakeaddress' }
+    return { publicAddress: 'fakeaddress', nativeAmount: '0' }
```
swansontec, line 268: The unit test needs an update now.

swansontec (approved):
Approved with one optional fixup.

PR #491, opened 2022-12-20
Clean swap quotes before logging

paullinator:
Prevents circular reference error due to wallets

---
  - https://app.asana.com/0/0/1203580495691905
  - https://app.asana.com/0/0/1203583134187240

### DIFF
--- a/src/core/swap/swap-api.js
+++ b/src/core/swap/swap-api.js
@@ -81,7 +81,9 @@ export async function fetchSwapQuote(
         })
         .then(
           quote => {
-            log.warn(`${pluginId} gave swap quote:`, quote)
+            const { fromWallet, toWallet, ...request } = quote.request ?? {}
+            const cleaned = { ...quote, request }
+            log.warn(`${pluginId} gave swap quote:`, cleaned)
             return quote
           },
           error => {


src/core/swap/swap-api.js
```diff
@@ -94,7 +95,7 @@ export async function fetchSwapQuote(
   if (promises.length < 1) throw new Error('No swap providers enabled')
 
   // Wait for the results, with error handling:
-  return fuzzyTimeout(promises, 20000).then(
+  return fuzzyTimeout(promises, 120000).then(
```
swansontec, line 98: Do we want 2 minutes? Remember, `fuzzyTimeout` will wait as long as it needs for at least one success to come back (even if that takes forever). This timeout is for situations where we already have successful results, but there are still a few pending promises. In that case, `fuzzyTimeout` will return what it has so far and ditch the stragglers.

src/core/swap/swap-api.js
```diff
@@ -81,7 +81,8 @@ export async function fetchSwapQuote(
         })
         .then(
           quote => {
-            log.warn(`${pluginId} gave swap quote:`, quote)
+            const cleaned = cleanQuote(quote)
+            log.warn(`${pluginId} gave swap quote:`, cleaned)
```
swansontec, line 85: The `cleanQuote` function messes up my TypeScript branch, since `Object` is more of a Flow thing. Let put the conversion inline, and use syntax that both types systems are happy with:

```ts
            const { fromWallet, toWallet, ...request } = quote.request
            const cleaned = { ...quote, request }
            log.warn(`${pluginId} gave swap quote:`, cleaned)
```

I tested this edited version, and it makes TypeScript happy too.

swansontec (changes requested):

src/core/swap/swap-api.js
paullinator: ack. That snuck in as it was for debugging. will remove

src/core/swap/swap-api.js
paullinator: Done. I added `= quote.request ?? {}` since anyone using a new core but with old exchange plugins might get blank `request` objects.

swansontec: Excellent

swansontec (approved):

PR #492, opened 2022-12-21
Add a new `alwaysEnabledTokenIds` API

swansontec:
These tokens will add to the per-wallet enabled token list, so the engine will always check the balances on these other currencies.

---
  - https://app.asana.com/0/0/1203582470755080


Jon-edge (approved):
Makes sense to me, thanks for the informative PR!

PR #493, opened 2022-12-23
Accelerate Transaction API

samholmes:
---
  - https://app.asana.com/0/0/1203510945643156

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.js
+++ b/src/core/currency/wallet/currency-wallet-api.js
@@ -521,6 +521,12 @@ export function makeCurrencyWalletApi(
       return engine.sweepPrivateKeys(spendInfo)
     },
 
+    // Accelerating:
+    async accelerate(tx: EdgeTransaction): Promise<EdgeTransaction | null> {
+      if (engine.accelerate == null) return null
+      return await engine.accelerate(tx)
+    },
+
     // Staking:
     get stakingStatus(): EdgeStakingStatus {
       return input.props.walletState.stakingStatus
--- a/src/core/currency/wallet/currency-wallet-api.js
+++ b/src/core/currency/wallet/currency-wallet-api.js
@@ -103,25 +103,25 @@ export function makeCurrencyWalletApi(
     watch: watchMethod,
 
     // Data store:
+    get disklet(): Disklet {
+      return storageWalletApi.disklet
+    },
     get id(): string {
       return storageWalletApi.id
-    },
-    get type(): string {
-      return storageWalletApi.type
     },
     get keys(): JsonObject {
       lockdown()
       return storageWalletApi.keys
     },
-    publicWalletInfo,
-    get disklet(): Disklet {
-      return storageWalletApi.disklet
-    },
     get localDisklet(): Disklet {
       return storageWalletApi.localDisklet
     },
+    publicWalletInfo,
     async sync(): Promise<void> {
       await storageWalletApi.sync()
+    },
+    get type(): string {
+      return storageWalletApi.type
     },
 
     // Wallet keys:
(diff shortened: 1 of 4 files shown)


src/types/types.js
```diff
@@ -630,6 +630,9 @@ export type EdgeCurrencyEngine = {
     paymentProtocolUrl: string
   ) => Promise<EdgePaymentProtocolInfo>,
 
+  // Accelerating:
+  +accelerate: (tx: EdgeTransaction) => Promise<EdgeTransaction | null>,
```
swansontec, line 634: This should be `accelerate?:`, since most engines won't have it implemented yet.

src/core/currency/wallet/currency-wallet-api.js
```diff
@@ -521,6 +521,11 @@ export function makeCurrencyWalletApi(
       return engine.sweepPrivateKeys(spendInfo)
     },
 
+    // Accelerating:
+    async accelerate(tx: EdgeTransaction): Promise<EdgeTransaction | null> {
+      return await engine.accelerate(tx)
```
swansontec, line 526: `if (typeof engine.accelerate !== 'function') return null`

test/fake/fake-currency-plugin.js
```diff
@@ -314,6 +314,11 @@ class FakeCurrencyEngine {
   saveTx(transaction: EdgeTransaction): Promise<void> {
     return Promise.resolve()
   }
+
+  // Accelerating:
+  accelerate(transaction: EdgeTransaction): Promise<EdgeTransaction | null> {
```
swansontec, line 319: We should upgrade this repo to use async more consistently, but we started coding in here before that was an option. I would rather see `async accelerate() (...) {return null}` going forward.

src/types/types.js
```diff
@@ -796,6 +799,9 @@ export type EdgeCurrencyWallet = {
   +signTx: (tx: EdgeTransaction) => Promise<EdgeTransaction>,
   +sweepPrivateKeys: (edgeSpendInfo: EdgeSpendInfo) => Promise<EdgeTransaction>,
 
+  // Accelerating:
+  +accelerate: (tx: EdgeTransaction) => Promise<EdgeTransaction | null>,
```
swansontec, line 803: I wonder if we should return `undefined` instead of `null` if this doesn't apply. We don't use `null` in very many places within the app.

swansontec (changes requested):
I am trusting you on the first commit. I have some small tweaks on the accelerate API, but otherwise I like the direction.

src/types/types.js
samholmes: I always followed the idea that `undefined` means something wasn't present or available, whereas `null` is an actual return value with meaning which is like a null object, or an empty object (`typeof null === 'object'` afterall). The idea is that null packs a little different semantic then `undefined` and in this case it means the implementation must be forced to return `null` explicitly (removes the risk of implementation accidentally not covering a codepath and therefore an omition of a return statement is equivalent to `return undefined`).

swansontec: Sure, makes sense.

swansontec (approved):

src/types/types.js
```diff
@@ -301,7 +301,6 @@ export type EdgeCurrencyInfo = {
   // Chain information:
   canAdjustFees?: boolean, // Defaults to true
   canImportKeys?: boolean, // Defaults to false
-  canReplaceByFee?: boolean, // Defaults to false
```
swansontec, line 304: We need to keep this flag, since we will use it to select between “pending” and “this transaction may be cancelled” in the GUI. So, it is no longer about *doing* RBF, but rather about whether RBF might be done to us.

swansontec (changes requested):

src/types/types.js
samholmes: Are you saying that this flag is to signal to the GUI whether a particular wallet supports RBF on transactions that may be broadcast, and this is useful in order to detect this behavior in the GUI before sending a transaction? Whether a particular transaction is replaceable is still determined by the `accelerate()` method, however this field in the currency info can still provide the GUI with information about the capabilities of the particular blockchain underpinning the wallet.

swansontec: No, it's about *receiving* money, from other people. Do we trust the sender, or not?

samholmes: Dropped the commit which was removing the `canReplaceByFee` field

PR #495, opened 2023-01-03
Add prefer swap DEX options

paullinator:
---
  - https://app.asana.com/0/0/1203628050381038

### DIFF
--- a/src/core/swap/swap-api.js
+++ b/src/core/swap/swap-api.js
@@ -5,7 +5,6 @@ import { bridgifyObject } from 'yaob'
 
 import { upgradeCurrencyCode } from '../../types/type-helpers.js'
 import {
-  type EdgePluginMap,
   type EdgeSwapQuote,
   type EdgeSwapRequest,
   type EdgeSwapRequestOptions,
@@ -28,7 +27,7 @@ export async function fetchSwapQuote(
   request: EdgeSwapRequest,
   opts: EdgeSwapRequestOptions = {}
 ): Promise<EdgeSwapQuote> {
-  const { preferPluginId, disabled = {}, promoCodes = {} } = opts
+  const { disabled = {}, preferPluginId, promoCodes = {} } = opts
   const { log } = ai.props
 
   const account = ai.props.state.accounts[accountId]
@@ -99,7 +98,7 @@ export async function fetchSwapQuote(
   return fuzzyTimeout(promises, 20000).then(
     quotes => {
       // Find the cheapest price:
-      const bestQuote = pickBestQuote(quotes, preferPluginId, promoCodes)
+      const bestQuote = pickBestQuote(quotes, opts)
       log.warn(
         `${promises.length} swap quotes requested, ${quotes.length} resolved, picked ${bestQuote.pluginId}.`
       )
@@ -110,6 +109,9 @@ export async function fetchSwapQuote(
       }
       // $FlowFixMe - Here for backwards compatibility:
       bestQuote.request = request
+      // $FlowFixMe - Here for backwards compatibility:
+      bestQuote.swapInfo = swapPlugins[bestQuote.pluginId].swapInfo
+
       return bridgifyObject(bestQuote)
     },
     (errors: any[]) => {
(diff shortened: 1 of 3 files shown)


src/types/types.js
```diff
@@ -828,12 +828,15 @@ export type EdgeCurrencyWallet = {
...
 
+export type EdgeSwapPluginType = 'DEX' | 'CEX' | 'transfer'
+
 /**
  * Static data about a swap plugin.
  */
 export type EdgeSwapInfo = {
   +pluginId: string,
   +displayName: string,
+  +swapPluginType?: EdgeSwapPluginType,
```
swansontec, line 839: I think this should be `isDex`, since there could be multiple overlapping attributes going forward: `isTrustless`, `isAtomic`, `isFixedPrice`, or whatever the future brings.

src/types/types.js
```diff
@@ -881,6 +884,7 @@ export type EdgeSwapQuote = {
   +fromNativeAmount: string,
   +toNativeAmount: string,
   +networkFee: EdgeNetworkFee,
+  +swapPluginType?: EdgeSwapPluginType,
```
swansontec, line 887: Would it make more sense to just put the `EdgeSwapInfo` here, to provide complete information about the chosen plugin? Otherwise, the `pluginId` should be enough to go look this up, without copying the fields over.

src/core/swap/swap-api.js
```diff
@@ -123,16 +129,30 @@ export async function fetchSwapQuote(
...
   quotes: EdgeSwapQuote[],
   preferPluginId: string | void,
-  promoCodes: EdgePluginMap<string>
+  promoCodes: EdgePluginMap<string>,
+  preferType?: EdgeSwapPluginType
 ): EdgeSwapQuote {
   return quotes.reduce((a, b) => {
+    // Prioritize transfer plugin:
+    if (a.swapPluginType === 'transfer') return a
+    if (b.swapPluginType === 'transfer') return b
```
swansontec, line 142: The transfer plugin is a hack that never really worked right, and we already plan to replace it with a transfer button on the QR modal.

I would be OK with `if (a.pluginId === 'transfer')`, since that doesn't leak this deprecated / unwanted feature out into the public type definitions.

src/core/swap/swap-api.js
```diff
@@ -123,16 +129,30 @@ export async function fetchSwapQuote(
  * Picks the best quote out of the available choices.
  * Exported so we can unit-test it.
  */
+
 export function pickBestQuote(
   quotes: EdgeSwapQuote[],
   preferPluginId: string | void,
-  promoCodes: EdgePluginMap<string>
+  promoCodes: EdgePluginMap<string>,
+  preferType?: EdgeSwapPluginType
```
swansontec, line 137: Extra whitespace? Also, can we just pass the whole `EdgeSwapRequestOptions` object in, since we pretty much need all of it?

swansontec (changes requested):
This makes a lot of sense, but I think a few small tweaks would make it better.

src/types/types.js
paullinator: That's fair. I can make that change.

src/types/types.js
paullinator: I originally wanted that but I considered that the pluginType might actually be something that slightly mismatches the quote. ie 'hybrid' plugin type that can do 'DEX' or 'CEX' or a 'hybrid' swap if we were to route through Changenow then Uniswap. Admittedly this begs the question of whether the quote should use a name `EdgeSwapPluginType` vs `EdgeSwapType`. The EdgeSwapInfo is for helping the GUI display plugins while the quote is useful for the matching engine.

This is all super nuanced and maybe overthinking so if you think this flexibility is over-architecting, then I'm 100% fine just plopping the EdgeSwapInfo into the quote.

src/core/swap/swap-api.js
paullinator: Ok. I'll change to that. Even with the improved transfer GUI, I don't want to remove the transfer plugin as doing so could route same-kind asset transfers through a swap partner and incur unnecessary fees.

src/core/swap/swap-api.js
paullinator: Flow isn't working on my VScode 🥲. I'm having to manually format then hope `yarn flow` `yarn lint` doesnt catch anything. Can't wait for that TS conversion! 👍 

Sure. I didn't want to have to redo all unit tests causing extra diffs but this makes sense.

src/types/types.js
swansontec: Well, the main thing is that we already have `pluginId`, so if we add more stuff like hybrid routing, we would have to deprecate / edit / adjust the `pluginId` as well. So, we can rely on the `pluginId` / `pluginInfo` to give us the information for now, and then cross the hybrid bridge when we come to it. For instance, maybe we define `pluginId: "hybrid"` to mean "go look elsewhere for this information".

src/core/swap/swap-api.js
swansontec: You can also do `yarn fix` on the CLI to auto-format everything.

src/types/types.js
paullinator: Ok. i'll just add the swapInfo

swansontec (approved):

src/types/types.js
```diff
@@ -878,13 +878,13 @@ export type EdgeSwapApproveOptions = {
  * If a provider can satisfy a request, what is their price?
  */
 export type EdgeSwapQuote = {
+  +swapInfo: EdgeSwapInfo,
```
swansontec, line 881: Actually, wait! The core is responsible for making sure this exists, even if the plugins are outdated. This means we need to upgrade the quote on the way out the door, in the same way we add the request itself if that is missing.

swansontec (changes requested):
I realized something was missing when I started reviewing the plugin changes. We need to make sure the quote gets upgraded in case the plugins are outdated for some reason.

swansontec (approved):

PR #499, opened 2023-01-24
Add 'syncing' to confirmations API

samholmes:
---
  - https://app.asana.com/0/0/1203810079930677

### DIFF
--- a/src/core/currency/wallet/currency-wallet-callbacks.js
+++ b/src/core/currency/wallet/currency-wallet-callbacks.js
@@ -327,25 +327,26 @@ const validateConfirmations = (
   blockHeight: number,
   requiredConfirmations: number = 1 // Default confirmation rule is 1 block
 ): $PropertyType<EdgeTransaction, 'confirmations'> => {
-  // If tx block height is ≤0, this means it's not yet mined in a block,
-  // so block confirmations is 0 (unconfirmed) or -1 (dropped).
-  const blockConfirmations =
-    tx.blockHeight <= 0 ? tx.blockHeight : 1 + blockHeight - tx.blockHeight
-
-  /*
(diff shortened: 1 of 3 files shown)


src/core/currency/wallet/currency-wallet-callbacks.js
```diff
@@ -327,25 +327,26 @@ const validateConfirmations = (
...
-    blockConfirmations >= requiredConfirmations
-      ? 'confirmed'
-      : blockConfirmations === 0
-      ? 'unconfirmed'
-      : blockConfirmations < 0
-      ? 'dropped'
-      : blockConfirmations
-
-  return confirmations
+  // If the transaction has a blockHeight >0, then it has been minded in a block
```
swansontec, line 330: s/minded/mined/

test/core/currency/confirmations.test.js
```diff
@@ -0,0 +1,60 @@
...
+    const tx: any = { blockHeight: txBlockHeight }
+    expect(
+      validateConfirmations(tx, netBlockHeight, required),
+      `Expected tx with blockHeight of ${tx.blockHeight} to be ${expected} at network blockHeight ${netBlockHeight} with ${required} required confs`
+    ).equals(expected)
+  }
+  const loop = (start: number, end: number, fn: (v: number) => void) => {
+    for (let i = start; i <= end; ++i) {
+      fn(i)
+    }
```
swansontec, line 24: I'm not sure these abstractions help that much. Unit tests should be obvious & dumb. I don't think you need to fix / re-write these tests, since they work, but they are harder to read & understand like this.

swansontec (approved):
Only one spelling mistake. Otherwise this will work fine.

PR #500, opened 2023-01-25
Typescript

swansontec:
Ports the library to TypeScript, but with auto-generated Flow types.

---
  - https://app.asana.com/0/0/1203777084541156


scripts/make-types.ts
```diff
@@ -0,0 +1,52 @@
...
+    .replace(/\bunknown\b/g, 'mixed')
+    .replace(/\| undefined\b/g, '| void')
+    .replace(/: undefined\b/g, ': void')
+
+    // Builtin types:
+    .replace(/\b(\w+): ComponentType</g, '$1: React$ComponentType<')
+    .replace(/\bPartial<(\w+)>/g, '$Rest<$1, { ... }>')
+
+  return '// @flow\n\n' + fixed
+}
```
samholmes, line 30: Do we need to put back the shims?

src/core/account/account-init.ts
```diff
@@ -48,8 +49,8 @@ async function createChildLogin(
   const parentKit: LoginKit = {
     serverPath: kit.serverPath,
     server: kit.server,
-    login: { children: [kit.login] },
-    stash: { children: [kit.stash] },
+    login: { children: [kit.login as LoginTree] },
+    stash: { children: [kit.stash as LoginStash] },
```
samholmes, line 53: Are these suppose to be size-one vectors?

src/core/login/create.ts
```diff
@@ -167,6 +167,6 @@ export async function createLogin(
   await loginFetch(ai, 'POST', kit.serverPath, request)
 
   kit.stash.lastLogin = now
-  await saveStash(ai, kit.stash)
-  return kit.login
+  await saveStash(ai, kit.stash as LoginStash)
+  return kit.login as LoginTree
```
samholmes, line 171: This doesn't make sense to me. How are you able to make this assertion?

src/types/exports.ts
```diff
@@ -96,7 +96,9 @@ export interface EdgeFakeWorldProps extends CommonProps {
  * We don't want this library to depend on `@types/react`,
  * since that isn't relevant for our Node or browser builds.
  */
-type ComponentType<Props> = React$ComponentType<Props>
+type ComponentType<Props> = (
+  props: Props
+) => { type: any; props: any; key: string | number | null }
```
samholmes, line 101: Should we not use `React.ComponentType`; if not, why?

package.json
```diff
@@ -70,7 +70,7 @@
     "prepare": "husky install && npm-run-all clean -p build.*",
     "start": "webpack serve",
     "test": "mocha -r sucrase/register 'test/**/*.test.ts'",
-    "types": "# tsc",
+    "types": "tsc",
```
samholmes, line 73: Clever

src/core/account/account-reducer.ts
```diff
@@ -168,11 +168,11 @@ const accountInner = buildReducer<AccountState, RootAction, AccountNext>({
       keysLoaded ? ids.filter(id => walletInfos[id].hidden) : []
   ),
 
-  keysLoaded(state = false, action: RootAction): boolean {
+  keysLoaded(state = false, action): boolean {
```
samholmes, line 171: I like this kind of cleanup!

samholmes (approved):
This all look correct to me, but I had some questions which may be valid.

src/core/login/login.js
```diff
@@ -64,7 +64,7 @@ export function searchTree<T>(
  * and the `update` callback replaces that node.
  * The `clone` callback updates the `children` on the non-replaced nodes.
  */
-function updateTree<Node: { +children?: any[] }, Output>(
+function updateTree<Node extends { readonly children?: any[] }, Output>(
```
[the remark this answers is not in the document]
swansontec: We have to do them here so we get valid syntax. Otherwise ESLint will fail to auto-fix the source code as part of the conversion script.

scripts/make-types.ts
swansontec: The "Builtin types" section contains the only shims we cared about.

src/core/account/account-init.ts
swansontec: These are arrays with a single element, that we have to type-cast to get rid of the `Partial<>`.

src/core/login/create.ts
swansontec: We are getting rid of the `Partial<>`, since we happen to know that these have all the stuff that should be there. Sorry, it's old code from before we had good type-checking, so things are a little looser here than they should be.

src/types/exports.ts
swansontec: We don't want to make `@types/react` a dependency, since that isn't relevant for the Web or Node targets.

PR #501, opened 2023-01-31
Add support for token activation

paullinator:
- Add unactivatedTokenIds property to EdgeCurrencyWallet
- Add optional tokenMap to EdgeCurrencyInfo so newer plugins can declare it to avoid legacy EdgeMetaToken[] getting auto converted to the tokenMap in edge-currency-accountbased makeOuterPlugin()

---
  - https://app.asana.com/0/0/1203860324833440

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -328,9 +328,11 @@ export interface EdgeMetadata {
   amountFiat?: number
 }
 
-export interface EdgeNetworkFee {
-  readonly currencyCode: string
+// Would prefer a better name than EdgeNetworkFee2 but can't think of one
+export interface EdgeNetworkFee2 {
   readonly nativeAmount: string
+  readonly currencyPluginId: string
+  readonly tokenId?: string
 }
 
 export interface EdgeTxSwap {
@@ -552,6 +554,24 @@ export type EdgeGetReceiveAddressOptions = EdgeCurrencyCodeOptions & {
   forceIndex?: number
 }
 
+export interface EdgeEngineGetActivationAssetsOptions {
+  // All wallets in the users account. This allows the engine to choose
+  // which wallets can fulfill this activation request
+  currencyWallets: { [walletId: string]: EdgeCurrencyWallet }
+
+  // If null, activate parent wallet:
+  activateTokenIds?: string[]
+}
+
+export interface EdgeEngineActivationOptions {
+  // If null, activate parent wallet:
+  activateTokenIds?: string[]
+
+  // Wallet if the user is paying with a different currency:
+  paymentWallet?: EdgeCurrencyWallet
+  paymentTokenId?: string
+}
+
 // engine --------------------------------------------------------------
 
 export interface EdgeCurrencyEngineCallbacks {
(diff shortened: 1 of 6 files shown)


src/types/types.ts
```diff
@@ -290,6 +295,7 @@ export interface EdgeCurrencyInfo {
   // Native token information:
   currencyCode: string
   denominations: EdgeDenomination[]
+  tokenMap?: EdgeTokenMap
```
swansontec, line 298: Tokens have to move from `EdgeCurrencyInfo` to `EdgeCurrencyConfig`, because that allows custom tokens to get equal treatment. The GUI had a ton of places where people used `currencyInfo.metaTokens` and completely forgot about custom tokens, so we absolutely do *not* want this here. The temptation to write a bug is too strong.

For plugins to list their builtin tokens, the new home is `EdgeCurrencyPlugin.getBuiltinTokens`. This will allow us to achieve another goal, which is to use the info server to update token lists.

src/types/types.ts
```diff
@@ -230,6 +230,11 @@ export interface EdgeToken {
   networkLocation: JsonObject | undefined
 }
 
+export interface EdgeAsset {
+  pluginId: string
+  tokenId?: string
+}
```
swansontec, line 236: I would rather just put `pluginId` and `tokenId` right in the quote / network fee / wherever. This is how the rest of the core API does it, and so far it has felt pretty seamless.

src/types/types.ts
paullinator: The issue I ran into that motivated this was `makeOuterPlugin()` in accountbased that would ALWAYS lowercase the contract address when converting to EdgeTokenMap. 

https://github.com/EdgeApp/edge-currency-accountbased/blob/master/src/common/innerPlugin.ts#L78

I needed to prevent that. Since it seemed EdgeTokenMap was the new format, I used that as the trigger to prevent the conversion and leave the EdgeMetaTokens alone. How else would you suggest to prevent `makeOuterPlugin()` from running.

src/types/types.ts
paullinator: My original thought was to make this concept universal and replace the EdgeTokenId in the GUI since that's a misnomer. tokenId is used to refer to the single string and not the pluginId/tokenId combo.

src/types/types.ts
paullinator: Actually, check out this commit in accountbased. I think I figured it out

https://github.com/EdgeApp/edge-currency-accountbased/pull/502/commits/63c8299e7cbc8f740f82479243fb755c6820ccdd

swansontec: The lowercase behavior is intentional and necessary for EVM chains. Rather than do a big refactor, Mathew and I agreed to do a "quick & dirty" upgrade as part of the bundle splitting. Now that build changes are done, I think we need to revisit the hacks with the design Matthew and I really wanted, but didn't have time to do.

src/types/types.ts
swansontec: Yes, I agree that `EdgeTokenId` should be renamed to `EdgeAsset` in the GUI. However, that doesn't mean we need to bring that type over here.

swansontec (approved):

swansontec (approved):

PR #502, opened 2023-02-14
Allow the info server to return dark & light logos

swansontec:
---
  - https://app.asana.com/0/0/1203971495274663

### DIFF
--- a/src/core/account/lobby-api.ts
+++ b/src/core/account/lobby-api.ts
@@ -1,5 +1,5 @@
-import { uncleaner } from 'cleaners'
-import { bridgifyObject } from 'yaob'
+import { asObject, asOptional, asString, uncleaner } from 'cleaners'
+import { bridgifyObject, close } from 'yaob'
 
 import { EdgeLobbyRequest } from '../../types/server-types'
 import { EdgeLobby, EdgeLoginRequest } from '../../types/types'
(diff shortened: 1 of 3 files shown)


src/core/account/lobby-api.ts
```diff
@@ -135,3 +136,32 @@ export async function makeLobbyApi(
...
+  lobbyJson: EdgeLobbyRequest,
+  appId: string
+): Promise<EdgeLoginRequest> {
+  const info = await fetchAppIdInfo(ai, appId)
+
+  // Make the API:
+  const out: EdgeLoginRequest = {
+    appId,
+    displayName: info.appName,
+    displayImageUrl: info.imageUrl ?? info.lightImageUrl ?? info.darkImageUrl,
```
paullinator, line 153: Shouldn't we assume `displayImageUrl` is a lightmode image since that's how it was used prior? Hence I would suggest

```
    displayName: info.appName,
    displayImageLightUrl: info.imageUrl ?? info.lightImageUrl,
    displayImageDarkUrl: info.darkImageUrl,
    displayImageLightUrl: info.lightImageUrl ?? info.imageUrl
```

This would make any old apps that only use `displayImageUrl` on a light background continue to always work even if new appIds are added that only support `displayImageDarkUrl` and `displayImageLightUrl`

I would also be ok ditching `displayImageLightUrl` and assume `displayImageLightUrl` is light

paullinator (changes requested):

paullinator (approved):

PR #506, opened 2023-02-25
Fix denom bug with built in tokens

paullinator:
Have getCurrencyMultiplier also accept builtInTokens retrieved via getBuiltInTokens()

Also change getCurrencyMultiplier to only take one plugin since that's all it ever did. And scanning multiple plugins for a currencyCode is incorrect

---
  - https://app.asana.com/0/0/1204057267024831

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -159,9 +159,10 @@ export function makeCurrencyWalletApi(
       denominatedAmount: string,
       currencyCode: string
     ): Promise<string> {
+      const plugin = input.props.state.plugins.currency[pluginId]
       const multiplier = getCurrencyMultiplier(
-        { [pluginId]: input.props.state.plugins.currency[pluginId] },
-        input.props.state.accounts[accountId].customTokens[pluginId],
+        plugin,
+        input.props.state.accounts[accountId].allTokens[pluginId],
         currencyCode
       )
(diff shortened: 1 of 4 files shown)


src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -159,8 +159,14 @@ export function makeCurrencyWalletApi(
       denominatedAmount: string,
       currencyCode: string
     ): Promise<string> {
+      const plugin = input.props.state.plugins.currency[pluginId]
```
swansontec, line 162: We should not be calling `getBuiltinTokens` here, since this has already been done once at load-time. Instead, use `input.props.state.accounts[accountId].allTokens[pluginId]`. That gives you both custom tokens and builtin tokens merged into a single `EdgeTokenMap`. It automatically updates as the user edits custom tokens.

Then you can just change the signature to `getCurrencyMultiplier(currencyInfo: EdgeCurrencyInfo, allTokens: EdgeTokenMap, currencyCode: string): string`. We don't need to worry about metaTokens anymore, since it's redundant with `builtinTokens` / `allTokens` now.

swansontec (changes requested):

src/core/currency/wallet/currency-wallet-api.ts
paullinator: 👍

swansontec (approved):

PR #508, opened 2023-03-01
Pass private keys to privileged engine methods

samholmes:
---
  - https://app.asana.com/0/0/1204056629741205
  - https://app.asana.com/0/0/1204207855596351

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -514,7 +514,9 @@ export function makeCurrencyWalletApi(
       )
     },
     async signTx(tx: EdgeTransaction): Promise<EdgeTransaction> {
-      return await engine.signTx(tx)
+      const privateKeys = walletInfo.keys
+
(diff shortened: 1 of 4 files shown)


swansontec (approved):
This was much simpler than I expected it to be! Let's hold off on merging until we have clarity on Monero and ZCash, though. More changes may be needed for those guys.

swansontec (changes requested):
Some small changes. Plus maybe the `opts: { privateKeys: {} }` thing we talked about. I would prefer "private" over "unsafe", since that's the term everybody uses in the crypto space.

swansontec (changes requested):
This is really close to being perfect! I am super happy with the way it's coming together. Just a few things left to look at.

samholmes:
Make `privateKeys: object` a parameter for `signTx`. Also make `privateKeys: object` type for options argument.

swansontec (approved):
This will work. I see that you skipped the one optional error-message change, but otherwise this is what we want.

PR #524, opened 2023-04-24
Log swap providers that time out

swansontec:
### CHANGELOG

- added: Log swap providers that time out

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204464475538412


samholmes (approved):
Makes sense

PR #527, opened 2023-05-04
Update unit tests

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Description

Stop using deprecated methods in our unit tests.

---
  - https://app.asana.com/0/0/1204527258277760

### DIFF
--- a/src/core/fake/fake-world.ts
+++ b/src/core/fake/fake-world.ts
@@ -144,11 +144,14 @@ export function makeFakeWorld(
       if (login == null) throw new Error(`Cannot find user ${account.username}`)
 
       // Figure out which repos to use:
-      const syncKeys = account.allKeys
-        .filter(info => info.keys?.syncKey != null)
-        .map(info =>
-          base16.stringify(base64.parse(info.keys.syncKey)).toLowerCase()
+      const syncKeys: string[] = []
+      for (const info of account.allKeys) {
+        const keys = await account.getRawPrivateKey(info.id)
+        if (keys.syncKey == null) continue
+        syncKeys.push(
+          base16.stringify(base64.parse(keys.syncKey)).toLowerCase()
         )
+      }
       const repos: EdgeFakeUser['repos'] = {}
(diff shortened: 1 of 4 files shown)


test/core/currency/wallet/currency-wallet.test.ts
```diff
@@ -210,28 +211,25 @@ describe('currency wallets', function () {
...
-    log.assert('customTokens changed')
+      displayName: 'Custom Token',
+      denominations: [{ multiplier: '1000', name: 'CUSTOM' }],
+      networkLocation: {
+        contractAddress:
+          '0X7CD5885327FD60E825D67D32F9D22B018227A208AA3C4819DA15B36B5D5869D3'
+      }
+    }
 
+    await config.addCustomToken(customToken)
```
samholmes, line 229: The customToken parameter here looks different from before. How is the networkLocation able to change without implementation change; are both network locations values cleaned to the same object in the assertion?

src/core/fake/fake-world.ts
```diff
@@ -144,11 +144,14 @@ export function makeFakeWorld(
...
 
       // Figure out which repos to use:
-      const syncKeys = account.allKeys
-        .filter(info => info.keys?.syncKey != null)
-        .map(info =>
-          base16.stringify(base64.parse(info.keys.syncKey)).toLowerCase()
+      const syncKeys: string[] = []
+      for (const info of account.allKeys) {
+        const keys = await account.getRawPrivateKey(info.id)
+        if (keys?.syncKey == null) continue
```
samholmes, line 150: No need for `?` now, I assume.

test/core/currency/wallet/currency-wallet.test.ts
swansontec: No, these are different functions. The `EdgeCurrencyConfig` version takes an `EdgeToken`, but the `EdgeCurrencyWallet` one takes an `EdgeMetaToken`, so the format has to change.

src/core/fake/fake-world.ts
swansontec: Good catch.

PR #528, opened 2023-05-05
Call setupNewTxMetadata for every new unknown transaction

peachbits:
The previous logic assumed that a send always originated from an Edge account and setupNewTxMetadata was already called during saveTx. This isn't the case for sends that are broadcast outside of the app and first seen through a transaction query

---
  - https://app.asana.com/0/0/1204535089316834


peachbits:
This is not the correct fix for missing transactions. We don't want to automatically create metadata files for sends because of a possible race condition with the engine. Instead, the transaction should appear even without a file present.

PR #550, opened 2023-07-25
Assume all fetch exceptions are potentially caused by CORS failure

samholmes:
### CHANGELOG

- Fixed: Fallback to CORS-safe fetch functions on all errors to fix inconsistency with error messages across platforms.

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205093858492115

### DIFF
--- a/src/io/react-native/react-native-worker.ts
+++ b/src/io/react-native/react-native-worker.ts
@@ -21,10 +21,14 @@ import { ClientIo, WorkerApi } from './react-native-types'
 
 // Hard-coded CORS proxy server
 const PROXY_SERVER_URLS = ['https://cors1.edge.app', 'https://cors2.edge.app']
+// Only try CORS proxy/bridge techniques up to 5 times
+const MAX_CORS_FAILURE_COUNT = 5
 // A map of domains that failed CORS and succeeded via the CORS proxy server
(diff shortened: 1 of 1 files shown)


src/io/react-native/react-native-worker.ts
```diff
@@ -21,10 +21,14 @@ import { ClientIo, WorkerApi } from './react-native-types'
 
 // Hard-coded CORS proxy server
 const PROXY_SERVER_URLS = ['https://cors1.edge.app', 'https://cors2.edge.app']
+// Only try CORS proxy/bridge techniques up to 5 times
+const MAX_CORS_FAILURE_COUNT = 5
 // A map of domains that failed CORS and succeeded via the CORS proxy server
 const hostnameProxyWhitelist = new Map<string, true>()
 // A map of domains that failed CORS and failed via the CORS proxy server and succeeded via native fetch
 const hostnameBridgeProxyWhitelist = new Map<string, true>()
+// A map of domains that all CORS techniques and should not re-attempt CORS techniques
```
paullinator, line 30: Seems like a mis typed comment. Did this mean to say

// A map of domains that failed all CORS techniques and should not re-attempt CORS techniques

paullinator (approved):
Just request one small fix to your comment

PR #553, opened 2023-07-28
Restore fetch behavior for security

swansontec:
We cannot send any secrets or credentials through the CORS proxy servers, since those aren't a secure environment.

Therefore, we need to restore the distinction between the two `fetch` methods. However, since we *do* have bouncer servers, we can finally offer the `fetchCors` methods on the other platforms.

### CHANGELOG

- changed: Remove `fetch` fallback logic. No proxy servers will be used.
- changed: The `fetchCors` method is no longer deprecated. Use this if CORS might be an issue. Do *not* use this for any secrets or credentials.

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205163372569911

### DIFF
--- a/src/io/browser/browser-io.ts
+++ b/src/io/browser/browser-io.ts
@@ -2,6 +2,16 @@ import { makeLocalStorageDisklet } from 'disklet'
 
 import { EdgeFetchOptions, EdgeFetchResponse, EdgeIo } from '../../types/types'
 import { scrypt } from '../../util/crypto/scrypt'
+import { fetchCorsProxy } from '../fetch-cors-proxy'
+
+// Only try CORS proxy/bridge techniques up to 5 times
+const MAX_CORS_FAILURE_COUNT = 5
+
+// A map of domains that failed CORS and succeeded via the CORS proxy server
+const hostnameProxyWhitelist = new Set<string>()
+
+// A map of domains that failed all CORS techniques and should not re-attempt CORS techniques
+const hostnameCorsProxyBlacklist = new Map<string, number>()
 
 /**
  * Extracts the io functions we need from the browser.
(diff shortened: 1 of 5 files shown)


src/io/browser/browser-io.ts
```diff
@@ -31,6 +35,32 @@ export function makeBrowserIo(): EdgeIo {
...
+      if (hostnameProxyWhitelist.has(hostname)) {
+        return await fetchCorsProxy(uri, opts)
+      }
+
+      try {
+        // Attempt regular fetch:
+        return await window.fetch(uri, opts)
+      } catch (error) {
+        // Fallback to edge-core-proxy server if this is a CORS error:
+        if (String(error) === 'TypeError: Load failed') {
```
samholmes, line 55: Not all errors messages are the same across all platforms, so we should assume all errors are potentially CORS related and try the cors proxy fetch.

src/io/browser/browser-io.ts
```diff
@@ -2,6 +2,10 @@ import { makeLocalStorageDisklet } from 'disklet'
 
 import { EdgeFetchOptions, EdgeFetchResponse, EdgeIo } from '../../types/types'
 import { scrypt } from '../../util/crypto/scrypt'
+import { fetchCorsProxy } from '../fetch-cors-proxy'
+
+// A map of domains that failed CORS and succeeded via the CORS proxy server
+const hostnameProxyWhitelist = new Set<string>()
```
samholmes, line 8: `Set` is a good idea.

samholmes (changes requested):

src/io/browser/browser-io.ts
samholmes: Also consider a Map of the failure counts to assume errors aren't to do with CORS

samholmes (approved):

PR #554, opened 2023-07-31
Throw an error if proxy server returns teapot error

peachbits:
This means the proxy server recognized it shouldn't handle the request and the core should try the next fallback

---
  - https://app.asana.com/0/0/1205166003289267

### DIFF
--- a/src/io/react-native/react-native-worker.ts
+++ b/src/io/react-native/react-native-worker.ts
@@ -12,7 +12,13 @@ import {
   makeContext,
   makeFakeWorld
 } from '../../core/core'
-import { EdgeFetchOptions, EdgeFetchResponse, EdgeIo } from '../../types/types'
+import {
+  asMaybeNetworkError,
+  EdgeFetchOptions,
+  EdgeFetchResponse,
+  EdgeIo,
+  NetworkError
+} from '../../types/types'
 import { fetchCorsProxy } from '../fetch-cors-proxy'
 import { hideProperties } from '../hidden-properties'
 import { makeNativeBridge } from './native-bridge'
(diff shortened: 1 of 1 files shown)


src/io/react-native/react-native-worker.ts
```diff
@@ -193,10 +195,14 @@ async function makeIo(clientIo: ClientIo): Promise<EdgeIo> {
       if (doFetchCors) {
         try {
           const response = await fetchCorsProxy(uri, opts)
+          if (response.status === 418) {
+            throw new Error(PROXY_ERROR)
```
swansontec, line 199: The core already defines a `NetworkError` type. Just throw that.

src/io/react-native/react-native-worker.ts
```diff
@@ -193,10 +195,14 @@ async function makeIo(clientIo: ClientIo): Promise<EdgeIo> {
...
+          }
           hostnameProxyWhitelist.add(hostname)
           return response
         } catch (error) {
-          if (errorToThrow == null) errorToThrow = error
+          if (errorToThrow == null && String(error) !== PROXY_ERROR)
```
swansontec, line 204: Here you can use `asMaybeNetworkError` to see if the error thrown is from the CORS proxy.

swansontec (changes requested):

swansontec (approved):
I fixed it for ya.

src/io/react-native/react-native-worker.ts
```diff
@@ -196,12 +200,12 @@ async function makeIo(clientIo: ClientIo): Promise<EdgeIo> {
...
           const response = await fetchCorsProxy(uri, opts)
           if (response.status === 418) {
-            throw new Error(PROXY_ERROR)
+            throw new NetworkError()
           }
           hostnameProxyWhitelist.add(hostname)
           return response
         } catch (error) {
-          if (errorToThrow == null && String(error) !== PROXY_ERROR)
+          if (errorToThrow == null && asMaybeNetworkError(error) != null)
```
peachbits, line 208: Nitpick, I think this should be:

asMaybeNetworkError(error) == null)

because the error we ultimately throw should originate from a fetch function and not our artificial teapot error

swansontec: Derp.

PR #556, opened 2023-08-08
Add isMaybePartial/maxFulfillmentTime to swap quote

paullinator:
### CHANGELOG

added: isMaybePartial/maxFulfillmentTime to EdgeSwapQuote

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205228135402239

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -1093,6 +1093,17 @@ export interface EdgeSwapQuote {
   readonly request: EdgeSwapRequest
 
   readonly isEstimate: boolean
+
+  /**
+   * This quote may be partially fulfilled with remaining source funds left
+   * in wallet
+   */
+  readonly canBePartial?: boolean
+
+  /**
+   * Maximum amount of time this quote will take to be fulfilled (in seconds)
+   */
+  readonly maxFulfillmentSeconds?: number
   readonly fromNativeAmount: string
   readonly toNativeAmount: string
   readonly networkFee: EdgeNetworkFee


src/types/types.ts
```diff
@@ -1093,6 +1093,17 @@ export interface EdgeSwapQuote {
...
+  /**
+   * This quote may be partially fulfilled with remaining source funds left
+   * in wallet
+   */
+  readonly isMaybePartial?: boolean
+
+  /**
+   * Maximum amount of time this quote will take to be fulfilled (in seconds)
+   */
+  readonly maxFulfillmentTime?: number
```
swansontec, line 1106: I would like `maxFulfillmentSeconds`, so the name caries its own units.

src/types/types.ts
```diff
@@ -1093,6 +1093,17 @@ export interface EdgeSwapQuote {
   readonly request: EdgeSwapRequest
 
   readonly isEstimate: boolean
+
+  /**
+   * This quote may be partially fulfilled with remaining source funds left
+   * in wallet
+   */
+  readonly isMaybePartial?: boolean
```
swansontec, line 1101: Would `canBePartial` or `allowsPartial` be better names? We use the word "maybe" a lot with cleaners, where it has a somewhat different connotation. Otherwise the name you have would be fine.

swansontec (approved):
Approved, but I would prefer the word "seconds" to appear in the fulfillment time if you agree.

src/types/types.ts
paullinator: `canBePartial` sounds good. will make the change.

src/types/types.ts
paullinator: sounds good

PR #557, opened 2023-08-15
Treat errors as `unknown`

swansontec:
### CHANGELOG

 none

### Dependencies

 none

### Description

This is an internal change (no changelog). We use `unknown` for errors, instead of `any`. 

---
  - https://app.asana.com/0/0/1205270928744917

### DIFF
--- a/src/types/exports.ts
+++ b/src/types/exports.ts
@@ -60,10 +60,15 @@ interface CommonProps {
   pluginUris?: string[]
 
   // Called if something goes wrong when starting the core:
+  // This will change to `(error: unknown) => void`
   onError?: (error: any) => unknown
 }
 
 export interface EdgeContextProps extends CommonProps {
+  /**
+   * Called once the core finishes loading.
+   * The return type will change to `=> void`
+   */
   onLoad: (context: EdgeContext) => unknown
 
   // EdgeFakeWorldOptions:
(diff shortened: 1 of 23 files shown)


src/types/exports.ts
```diff
@@ -60,10 +60,15 @@ interface CommonProps {
   pluginUris?: string[]
 
   // Called if something goes wrong when starting the core:
+  // This will change to `(error: unknown) => void`
```
samholmes, line 63: When?

src/types/types.ts
```diff
@@ -1588,7 +1588,7 @@ export interface EdgeUserInfo {
 
 export interface EdgeContextEvents {
   close: void
-  error: Error
+  error: any // Will change to `unknown`
```
samholmes, line 1591: Why change it to `any`?

samholmes (approved):
Approved but I had a couple of questions.

src/types/exports.ts
swansontec: This will be a breaking change, so whenever we bump the version number to 2.0.

src/types/types.ts
swansontec: We have no way to guarantee that the field will be an `Error` object. It could be an error or a string or `null` or whatever else gets passed to `throw` deep in some library. Since we can't make this promise, we should just be honest about that fact.

PR #558, opened 2023-08-16
Reduce yaob throttle to 50ms

paullinator:
### CHANGELOG

Speed up GUI interaction by adding yaob throttling of 50ms

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205274123045114

### DIFF
--- a/src/io/react-native/react-native-worker.ts
+++ b/src/io/react-native/react-native-worker.ts
@@ -22,7 +22,7 @@ import {
 import { fetchCorsProxy } from '../fetch-cors-proxy'
 import { hideProperties } from '../hidden-properties'
 import { makeNativeBridge } from './native-bridge'
-import { ClientIo, WorkerApi } from './react-native-types'
+import { ClientIo, WorkerApi, YAOB_THROTTLE_MS } from './react-native-types'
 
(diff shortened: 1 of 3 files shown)


src/io/react-native/react-native-worker.ts
```diff
@@ -36,6 +36,9 @@ const hostnameBridgeProxyWhitelist = new Set<string>()
 // A map of domains that failed all CORS techniques and should not re-attempt CORS techniques
 const hostnameCorsProxyBlacklist = new Map<string, number>()
 
+// Throttle YAOB updates
+const YAOB_THROTTLE_MS = 75
```
swansontec, line 40: You also need to pass this to src/io/react-native/yaob-callbacks.ts, on line 80. That is where the bridge lives going the other way.

Since react-native code can't import web code or vice-versa, we need to put this constant in react-native-types.ts, which is "common ground" between the two sides.

swansontec (changes requested):

paullinator:
I changed the throttle to 50ms to compensate for the roundtrip throttling. branch name is wrong now but  that will eventually go away.

swansontec (approved):

PR #562, opened 2023-09-04
Add a `ChallengeError` and related types

swansontec:
### CHANGELOG

- added: Add a new `ChallengeError` and related types.

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1204093435054224

### DIFF
--- a/src/core/login/login-fetch.ts
+++ b/src/core/login/login-fetch.ts
@@ -1,11 +1,13 @@
-import { uncleaner } from 'cleaners'
+import { asMaybe, uncleaner } from 'cleaners'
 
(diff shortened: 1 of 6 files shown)


samholmes (changes requested):
So far this looks good, but I assume it's not complete; where is the `ChallengeError` being instantiated and thrown? Is the login UI going to be responsible for constructing the `ChallengeError` type from the response? I'd say to that why not have the core throw the error type so that the login-ui can be responsible for just the UI and UX flow?

samholmes (approved):

PR #567, opened 2023-10-06
Add new EdgeTxInfo types

Jon-edge:
### CHANGELOG

- added: EdgeTxInfo types to tag known transaction types (swap, stake, etc)

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205412691849552

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -249,6 +249,39 @@ export interface EdgeMemo {
   memoName?: string
 }
 
+export interface EdgeAssetAmount {
+  pluginId: string
+  tokenId?: string
+  nativeAmount?: string
+}
+
+export type EdgeTxActionSwapType =
+  | 'swap'
+  | 'swapOrderPost'
+  | 'swapOrderFill'
+  | 'swapOrderCancel'
+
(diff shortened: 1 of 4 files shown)


src/types/types.ts
```diff
@@ -249,6 +249,28 @@ export interface EdgeMemo {
...
+export interface EdgeTxAsset {
+  pluginId: string
+  tokenId: string
+  nativeAmount: string
+}
+
+export interface EdgeTxInfoSwap {
+  type: 'swap' | 'swapOrderPost' | 'swapOrderFill'
+  orderId?: string
+  direction: 'from' | 'to'
```
Jon-edge, line 261: I question the need for this. In our exchange plugins this denotes which side is
the guaranteed amount - the payment or receive amount. In this API, we already
define a nativeAmount for both sides in EdgeTxAsset. We can just make
nativeAmount optional to denote an unknown swap destAsset amount for some order
with slippage.

swansontec (approved):
Approving this part, but we need to do more inner plumbing to get the data sent to the right place.

PR #574, opened 2023-11-30
Paul/saved action

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206044971437796

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -682,7 +682,7 @@ export function combineTxWithFile(
 
   // Copy the tx properties to the output:
   const out: EdgeTransaction = {
-    action: tx.action,
+    chainAction: tx.chainAction,
     blockHeight: tx.blockHeight,
     confirmations: tx.confirmations,
     currencyCode,
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -579,18 +579,36 @@ export function makeCurrencyWalletApi(
       fakeCallbacks.onTransactionsChanged([tx])
     },
 
+    async saveTxAction(opts): Promise<void> {
+      const { txid, tokenId, assetAction, savedAction } = opts
+      await setCurrencyWalletTxMetadata(
+        input,
+        txid,
+        tokenId,
+        fakeCallbacks,
+        undefined,
+        assetAction,
+        savedAction
+      )
+    },
+
     async saveTxMetadata(
       txid: string,
       currencyCode: string,
       metadata: EdgeMetadataChange
     ): Promise<void> {
+      const { tokenId = null } = upgradeCurrencyCode({
+        allTokens: input.props.state.accounts[accountId].allTokens[pluginId],
+        currencyInfo: plugin.currencyInfo,
+        currencyCode
+      })
       upgradeMetadata(input, metadata)
       await setCurrencyWalletTxMetadata(
         input,
         txid,
-        currencyCode,
-        metadata,
-        fakeCallbacks
+        tokenId,
+        fakeCallbacks,
+        metadata
       )
     },
 
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -40,7 +40,12 @@ import { toApiInput } from '../../root-pixie'
 import { makeStorageWalletApi } from '../../storage/storage-api'
 import { getCurrencyMultiplier } from '../currency-selectors'
 import { makeCurrencyWalletCallbacks } from './currency-wallet-callbacks'
-import { asEdgeTxSwap, TransactionFile } from './currency-wallet-cleaners'
+import {
+  asEdgeAssetAction,
+  asEdgeTxAction,
+  asEdgeTxSwap,
+  TransactionFile
+} from './currency-wallet-cleaners'
 import { dateFilter, searchStringFilter } from './currency-wallet-export'
 import {
   loadTxFiles,
@@ -481,14 +486,16 @@ export function makeCurrencyWalletApi(
     async makeSpend(spendInfo: EdgeSpendInfo): Promise<EdgeTransaction> {
       spendInfo = upgradeMemos(spendInfo, plugin.currencyInfo)
       const {
+        assetAction,
         customNetworkFee,
+        memos,
         metadata,
         networkFeeOption = 'standard',
         noUnconfirmed = false,
         otherParams,
         pendingTxs,
         rbfTxid,
-        memos,
+        savedAction,
         skipChecks,
         spendTargets = [],
         swapData
@@ -559,6 +566,8 @@ export function makeCurrencyWalletApi(
       tx.spendTargets = savedTargets
       if (metadata != null) tx.metadata = metadata
       if (swapData != null) tx.swapData = asEdgeTxSwap(swapData)
+      if (savedAction != null) tx.savedAction = asEdgeTxAction(savedAction)
+      if (assetAction != null) tx.assetAction = asEdgeAssetAction(assetAction)
       if (input.props.state.login.deviceDescription != null)
         tx.deviceDescription = input.props.state.login.deviceDescription
 
@@ -683,6 +692,7 @@ export function combineTxWithFile(
   // Copy the tx properties to the output:
   const out: EdgeTransaction = {
     chainAction: tx.chainAction,
+    chainAssetAction: tx.chainAssetAction.get(tokenId),
     blockHeight: tx.blockHeight,
     confirmations: tx.confirmations,
     currencyCode,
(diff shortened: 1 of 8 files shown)


src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -701,6 +704,18 @@ export function combineTxWithFile(
...
+      file.tokens[PARENT_TOKEN_ID],
+      file.tokens[tokenId ?? PARENT_TOKEN_ID]
+    )
+
+    if (mergedTokens.savedAction != null) {
+      out.savedAction = {
+        ...out.savedAction,
+        ...mergedTokens.savedAction
+      }
+    }
```
swansontec, line 743: I think this whole block could just be `out.savedAction = mergedTokens.savedAction`. There is no way `out.savedAction` can be anything other than `undefined`, if you follow the logic in this function.

I see similar logic above for the metadata, which is probably also a mistake. At least with metadata, though, you can always merge two structures together and get a valid metadata. With transaction actions, though, merging two together will usually produce an invalid value.

src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -560,6 +561,31 @@ export function makeCurrencyWalletApi(
...
+    ): Promise<void> {
+      const { accountApi } = input.props.output.accounts[accountId]
+      const { allTokens, currencyInfo } = accountApi.currencyConfig[pluginId]
+
+      const { currencyCode } = upgradeCurrencyCode({
+        allTokens,
+        currencyInfo,
+        currencyCode: undefined,
+        tokenId: tokenId ?? undefined
+      })
```
swansontec, line 577: Try `const { currencyCode } = tokenId == null ? currencyInfo : allTokens[tokenId]`. It's shorter and simpler.

The `upgradeCurrencyCode` function is designed to deal with backwards-compatibility issues, and since this is brand new code, it seems weird to use it.

test/core/currency/wallet/currency-wallet.test.ts
```diff
@@ -544,6 +558,18 @@ describe('currency wallets', function () {
       exchangeAmount: { 'iso:USD': 0.75 },
       ...metadata
     })
+    expect(txs[0].savedAction).deep.equals({
```
swansontec, line 582: You could do `expect(txs[0].savedAction).deep.equals(savedAction)`

src/core/currency/wallet/currency-wallet-cleaners.ts
```diff
@@ -246,7 +254,37 @@ export const asEdgeTxActionStake = asObject({
...
+  'buy',
+  'sell',
+  'sellNetworkFee'
+)
+
+export const asEdgeTxActionFiat = asObject<EdgeTxActionFiat>({
+  type: asEdgeTxActionFiatType,
+
+  orderId: asString,
+  walletId: asString,
```
swansontec, line 281: Why is there a walletId in here? We are already attached to an `EdgeTransaction`, and that has a `walletId` on it.

src/core/currency/wallet/currency-wallet-cleaners.ts
```diff
@@ -246,7 +254,37 @@ export const asEdgeTxActionStake = asObject({
   stakeAssets: asArray(asEdgeAssetAmount)
 })
 
-export const asEdgeTxAction = asEither(asEdgeTxActionSwap, asEdgeTxActionStake)
+export const asEdgeTxActionFiatType: Cleaner<EdgeTxActionFiatType> = asValue(
+  'buy',
+  'sell',
+  'sellNetworkFee'
```
swansontec, line 274: OK, so with this design, we need to save two copies of the action, duplicating the orderId and all other fields. One copy goes with the parent currency as a fee, and one goes with the token. The only difference between the two copies is the `type` field.

You already have a `tokenId` inside the action itself, so there is no need for this duplication. The GUI can easily see which side is the fee and which side is the sell by using `tx.tokenId === savedAction.cryptoAsset.tokenId`.

So, I really think the whole concept of saving the action per-currency is a mistake. There should be one action per transaction, saved at the root level.

When we first discussed this, I trusted your assertion that the actions needed to be saved per-token. Now that I'm seeing what the action looks like, I think that's a deeply flawed assumption.

swansontec (changes requested):

src/core/currency/wallet/currency-wallet-cleaners.ts
```diff
@@ -193,6 +207,47 @@ export function asIntegerString(raw: unknown): string {
 // file cleaners
 // ---------------------------------------------------------------------
 
+export const asNumeralString = (raw: any): string => {
+  if (typeof raw !== 'string') throw new Error('Not type string')
+  if (isNaN(Number(raw))) throw new Error('Does not parse as a number')
+  return raw
+}
```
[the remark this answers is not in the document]
paullinator: The intent was decimal support although unneeded for its current use. Once Sam merges his biggystring PR we'll have full scientific notation support. For now I'll change to `asIntegerString`

src/core/currency/wallet/currency-wallet-cleaners.ts
paullinator: In the case of a fiat transaction this might be simple, but consider a DEX swap on Ripple. You have an action to create an order which pays XRP, the outgoing send which is considered a 'swap' which could pay USD.bs, then the receive which is the fulfillment that receives USD.gh. Do you really want the GUI to have to figure that all out.

As well the plugins report transactions per asset as they send individual EdgeTransactions. Now while they might not use `savedAction` but instead `action`, they would still be putting the action info per asset which we don't want to conflict.

src/core/currency/wallet/currency-wallet-cleaners.ts
paullinator: That's fair. I'll remove.

src/core/currency/wallet/currency-wallet-reducer.ts
```diff
@@ -52,9 +53,13 @@ export interface MergedTransaction {
   ourReceiveAddresses: string[]
   signedTx: string
   txid: string
-
-  nativeAmount: { [currencyCode: string]: string }
-  networkFee: { [currencyCode: string]: string }
+  tokenId: string | null
```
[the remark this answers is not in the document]
paullinator: Ok. Agreed. But such a change is a bit outside the scope of this PR. Can we punt that to another task. The tokenId here at least mirrors the currencyCode use and they can both be fixed simultaneously when we agree on the method.

src/core/currency/wallet/currency-wallet-cleaners.ts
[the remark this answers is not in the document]
swansontec: Even if biggystring supports full exponent syntax, I think it would still make sense to require integers here - these are native amounts, so if a decimal point sneaks in, we have a bug.

src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -65,6 +69,8 @@ const fakeMetadata = {
   notes: ''
 }
 
+export const PARENT_TOKEN_ID = ''
+
```
swansontec, line 73: I wonder if this line should live in `currency-wallet-files.ts`, since it's related to the on-disk representation.

src/core/currency/wallet/currency-wallet-cleaners.ts
```diff
@@ -20,10 +33,10 @@ import { asJsonObject } from '../../../util/file-helpers'
  */
 export interface DiskMetadata {
   bizId?: number
-  category?: string
+  category?: string | null
   exchangeAmount: { [fiatCurrencyCode: string]: number }
-  name?: string
-  notes?: string
+  name?: string | null
+  notes?: string | null
```
swansontec, line 39: We don't actually want to store `null` on disk. The only reason we allow `null` in the user-facing type is to be able to delete fields. We need to revert this type definition, plus whatever changes are needed to the merging / updating code to make that work.

If could make sense to defined an `EdgeMetadataChange` type that includes `| null` for the fields that can be deleted, and revert the original `EdgeMetadata` type to how it was before. That makes it clear that metadata will never *contain* nulls. It's just a sigil for deleting fields.

src/core/currency/wallet/currency-wallet-cleaners.ts
```diff
@@ -175,10 +204,10 @@ export const asEdgeTxSwap = asObject<EdgeTxSwap>({
 
 const asDiskMetadata = asObject<DiskMetadata>({
   bizId: asOptional(asNumber),
-  category: asOptional(asString),
+  category: asNullable(asString),
   exchangeAmount: asOptional(asObject(asNumber), () => ({})),
-  name: asOptional(asString),
-  notes: asOptional(asString)
+  name: asNullable(asString),
+  notes: asNullable(asString)
```
swansontec, line 210: Same here. We don't want to save `null` on disk.

swansontec (changes requested):
Just some changes on the metadata stuff.

I don't really understand the new `savedAction` / `chainAction` / `assetAction` design. I see that `savedAction` / `assetAction` are per-transaction, which makes sense, and the new `assetAction` is per asset. What I don't understand is how the plugin / app coordinate on setting up `assetAction`.

paullinator:
> Just some changes on the metadata stuff.
> 
> I don't really understand the new `savedAction` / `chainAction` / `assetAction` design. I see that `savedAction` / `assetAction` are per-transaction, which makes sense, and the new `assetAction` is per asset. What I don't understand is how the plugin / app coordinate on setting up `assetAction`.

This works as follows. Plugins edit `chainAction` which is per tx and `chainAssetAction` which is per asset per tx

GUI edits `savedAction` which is per tx and `assetAction` which is per asset per tx. All four are needed unfortunately because of separation of chain vs gui and per asset vs per tx.

src/core/currency/wallet/currency-wallet-cleaners.ts
paullinator: Understood. I noticed this when testing. I'd rather not intro a new type just yet but I can ensure we don't save null on disk.

swansontec (approved):

PR #575, opened 2023-12-01
Create wallets

swansontec:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205970913826995

### DIFF
--- a/src/core/account/account-api.ts
+++ b/src/core/account/account-api.ts
@@ -30,11 +30,7 @@ import {
 import { base58 } from '../../util/encoding'
 import { getPublicWalletInfo } from '../currency/wallet/currency-wallet-pixie'
 import { makeExchangeCache } from '../exchange/exchange-api'
-import {
-  createCurrencyWallet,
-  makeKeysKit,
-  makeStorageKeyInfo
-} from '../login/keys'
+import { createCurrencyWallet, makeKeyInfo, makeKeysKit } from '../login/keys'
 import { applyKit } from '../login/login'
 import { deleteLogin } from '../login/login-delete'
 import { changeUsername } from '../login/login-username'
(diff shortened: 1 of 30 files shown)


test/fake/fake-swap-plugin.ts
```diff
@@ -16,19 +17,26 @@ const swapInfo: EdgeSwapInfo = {
...
   fetchSwapQuote(
     request: EdgeSwapRequest,
-    userSettings: JsonObject = {}
+    userSettings: object = {}
   ): Promise<EdgeSwapQuote> {
+    const cleanSettings = asFakeSwapSettings(userSettings)
+
     // We need KYC:
-    if (typeof userSettings.kycToken !== 'string') {
+    if (typeof cleanSettings.kycToken !== 'string') {
```
peachbits, line 39: I'm just a teeny bit annoyed you changed the checkSettings typeof kycToken comparison to a null check but not this one

src/core/account/account-api.ts
```diff
@@ -503,6 +504,29 @@ export function makeAccountApi(ai: ApiInput, accountId: string): EdgeAccount {
...
+    ): Promise<EdgeCurrencyWallet[]> {
+      const { login, loginTree } = accountState()
+
+      const walletInfos = await Promise.all(
+        createWallets.map(async opts => {
+          return await makeCurrencyWalletKeys(ai, opts.walletType, opts)
+        })
+      )
+      await applyKit(ai, loginTree, makeKeysKit(ai, login, walletInfos))
+      return await Promise.all(
```
peachbits, line 518: In a situation where some wallets are created and some aren't (failure for whatever reason) should this function still return an array of wallets and errors or throw if there's just a single error? The successful ones would still be present in the account even if this function doesn't return, presumably.

peachbits (approved):

test/fake/fake-swap-plugin.ts
swansontec: Derp! Good catch.

peachbits (approved):

PR #587, opened 2024-03-11
Fetch plugin data from info server

swansontec:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206639139959674


src/core/context/info-cache-file.ts
```diff
@@ -0,0 +1,19 @@
...
+
+export interface InfoCacheFile {
+  corePlugins?: EdgePluginMap<JsonObject>
+  syncServers?: string[]
+}
+
+export const INFO_CACHE_FILE_NAME = 'infoCache.json'
+
+export const asInfoCacheFile = asObject<InfoCacheFile>({
+  // date: asDate,
```
swansontec, line 14: I had considered storing the date we fetched this, but I guess it ended up not being needed - I should probably delete this.

src/core/currency/currency-pixie.ts
```diff
@@ -21,5 +23,28 @@ export const currency: TamePixie<RootProps> = combinePixies({
       walletState: props.state.currency.wallets[walletId],
       walletOutput: props.output.currency.wallets[walletId]
     })
-  )
+  ),
+
+  pluginUpdater(input: ApiInput) {
```
samholmes, line 28: When does does this pixie get invoked?

src/core/currency/currency-pixie.ts
```diff
@@ -21,5 +23,28 @@ export const currency: TamePixie<RootProps> = combinePixies({
...
+          plugin.updateInfoPayload != null &&
+          lastInfo != null &&
+          !matchJson(oldPayload, newPayload)
+        ) {
+          await plugin.updateInfoPayload(newPayload)
+        }
+      }
+      lastInfo = infoCache
+    }
+  }
```
samholmes, line 49: This don't call updateInfoPayload on the first invocation, so I'm guessing that the first invocation is right after initializing the plugin with the info payload?

src/types/types.ts
```diff
@@ -1311,8 +1315,9 @@ export interface EdgeSwapPlugin {
   readonly fetchSwapQuote: (
     request: EdgeSwapRequest,
     userSettings: JsonObject | undefined,
-    opts: { promoCode?: string }
+    opts: { infoPayload: JsonObject; promoCode?: string }
   ) => Promise<EdgeSwapQuote>
+  readonly updateInfoPayload?: (infoPayload: JsonObject) => Promise<void>
```
samholmes, line 1320: I thought were weren't going to have the updateInfoPayload method on the swap plugins?

samholmes (changes requested):

src/core/currency/currency-pixie.ts
swansontec: The pixie exists during the lifetime of the core, and the returned updater function gets invoked any time Redux changes.

src/core/currency/currency-pixie.ts
swansontec: Right, that was the intent. The plugins get the initial value at boot-up, so we only need to update them if something has changed since then.

src/core/currency/currency-pixie.ts
```diff
@@ -31,17 +31,22 @@ export const currency: TamePixie<RootProps> = combinePixies({
...
-          await plugin.updateInfoPayload(newPayload)
+      // Bail out quickly if nothing has changed:
+      if (lastInfo === infoCache) return
+
+      // Update plugins after the first run:
+      if (lastInfo != null) {
+        for (const pluginId of Object.keys(plugins.currency)) {
+          const plugin = plugins.currency[pluginId]
+          const newPayload = infoCache.corePlugins?.[pluginId] ?? {}
+          const oldPayload = lastInfo?.corePlugins?.[pluginId] ?? {}
```
samholmes, line 42: Unnecessary `?.` for `lastInfo?.`

src/util/match-json.ts
samholmes, line 1: Oof! I didn't know match-json.ts was missing from the first review. Good catch

samholmes (approved):
Approved with minor nit

PR #589, opened 2024-03-21
Add `'failed'` variant to EdgeTransaction `confirmations` API

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1206892270756769

### DIFF
--- a/src/core/currency/wallet/currency-wallet-callbacks.ts
+++ b/src/core/currency/wallet/currency-wallet-callbacks.ts
@@ -223,7 +223,7 @@ export function makeCurrencyWalletCallbacks(
                 input.props.walletState.currencyInfo
               const { height } = input.props.walletState
 
-              reduxTx.confirmations = validateConfirmations(
+              reduxTx.confirmations = determineConfirmations(
                 reduxTx,
                 height,
                 requiredConfirmations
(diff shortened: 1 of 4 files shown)


src/core/currency/wallet/currency-wallet-callbacks.ts
```diff
@@ -417,6 +411,17 @@ export function watchCurrencyWallet(input: CurrencyWalletInput): void {
   checkChangesLoop()
 }
 
+// DEPRECATE: After all currency plugins implement new Confirmations API
+// This function is used to determine if the core should determine the confirmations
+// field value for a transaction.
```
swansontec, line 416: Please use the JavaDoc style:, with a summary of what the function does and an explicit deprecation marker:

```ts
/** 
 * Returns true if the core needs to calculate the transaction's confirmation state.
 * 
 * @deprecated Remove once all currency plugins support the new confirmations API.
 */
```

When I do a major breaking change in the core, I always so a search for the `@deprecated` tag, so this will be helpful in actually getting the code deleted.

swansontec (approved):
Approved, but please fix the comment on the new utility function

swansontec (approved):

PR #598, opened 2024-06-07
New enableRbf field on EdgeSpendInfo

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1205841634636330
  - https://app.asana.com/0/0/1205841634636328

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- added: New `enableRbf` field on `EdgeSpendInfo`
+
 ## 2.6.1 (2024-05-23)
 
 - fixed: Add missing Swift import statement.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 ## Unreleased
 
 - added: New `enableRbf` field on `EdgeSpendInfo`
+- changed: Deprecated `rbfTxid` field on `EdgeSpendInfo`
 
 ## 2.6.1 (2024-05-23)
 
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,7 @@
 
 ## Unreleased
 
+- added: New `enableRbf` field on `EdgeSpendInfo`
```
swansontec, line 5: Missing newline.

src/types/types.ts
```diff
@@ -667,6 +667,8 @@ export interface EdgeSpendInfo {
   noUnconfirmed?: boolean
   networkFeeOption?: 'high' | 'standard' | 'low' | 'custom'
   customNetworkFee?: JsonObject // Some kind of currency-specific JSON
+  /** Enables/disables RBF for currencies where RBF is opt-in or optional. */
+  enableRbf?: boolean
```
swansontec, line 671: This would read better as: "Enables RBF on chains where RBF is optional".

Also, you need to edit currency-wallet-api around line 460 to actually pass this new field down to the engine. We filter the object so unknown fields won't pass through.

src/types/types.ts
```diff
@@ -670,6 +670,7 @@ export interface EdgeSpendInfo {
   /** Enables/disables RBF for currencies where RBF is opt-in or optional. */
   enableRbf?: boolean
   pendingTxs?: EdgeTransaction[]
+  /** @deprecated Use enableRbf instead */
```
swansontec, line 673: The `enableRbf` flag isn't right alternative. This comment should say "Use `EdgeCurrencyWallet.accelerate` instead"

swansontec (changes requested):

swansontec (approved):

PR #599, opened 2024-06-19
Add makeMemoryWallet method

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207435895725678
  - https://app.asana.com/0/0/1207585261612923

### DIFF
--- a/src/core/account/account-api.ts
+++ b/src/core/account/account-api.ts
@@ -17,6 +17,7 @@ import {
   EdgeGetActivationAssetsOptions,
   EdgeGetActivationAssetsResults,
   EdgeLobby,
+  EdgeMemoryWallet,
   EdgePendingVoucher,
   EdgePluginMap,
   EdgeResult,
@@ -59,6 +60,7 @@ import { changeWalletStates } from './account-files'
 import { AccountState } from './account-reducer'
 import { makeDataStoreApi } from './data-store-api'
 import { makeLobbyApi } from './lobby-api'
+import { makeMemoryWalletInner } from './memory-wallet'
 import { CurrencyConfig, SwapConfig } from './plugin-api'
 
 /**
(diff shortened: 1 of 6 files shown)


src/core/account/account-api.ts
```diff
@@ -498,6 +510,197 @@ export function makeAccountApi(ai: ApiInput, accountId: string): EdgeAccount {
...
+            syncNetworkTask.stop()
+          }
+        }
+        syncNetworkTask = makePeriodicTask(doNetworkSync, 10000, {
+          onError: error => {
+            ai.props.log.error(error)
+          }
+        })
+        syncNetworkTask.start({ wait: false })
+        await syncNetworkTask
```
swansontec, line 642: Also don't `await syncNetworkTask` here.

src/core/account/account-api.ts
```diff
@@ -498,6 +510,197 @@ export function makeAccountApi(ai: ApiInput, accountId: string): EdgeAccount {
...
+      const privateKeys = { ...keys }
+
+      const syncNetwork = async (): Promise<void> => {
+        try {
+          syncNetworkTask.start({ wait: false })
+          await syncNetworkTask
+        } catch (error) {
+          syncNetwork().finally(() => {})
+        }
+      }
```
swansontec, line 620: Get rid of this whole block, and just do `syncNetworkTask?.start({ wait: false })` directly in the `startEngine` method. There is no need to `await syncNetworkTask` because it's not a promise. Also, the start & stop methods can't throw - errors come out through the `onError` callback.

src/core/account/account-api.ts
```diff
@@ -498,6 +510,197 @@ export function makeAccountApi(ai: ApiInput, accountId: string): EdgeAccount {
...
+            tx,
+            unsafeBroadcastTx ? privateKeys : undefined
+          )
+        },
+        async saveTx() {},
+
+        async close() {
+          log.warn('killing memory wallet')
+          close(out)
+          await engine.killEngine()
```
swansontec, line 696: `syncNetworkTask?.stop()`

src/types/types.ts
```diff
@@ -905,11 +905,15 @@ export interface EdgeCurrencyEngineOptions {
   userSettings: JsonObject | undefined
 }
 
+export interface StartEngineOpts {
+  lightMode?: boolean
+}
```
swansontec, line 910: Please move this flag to `makeCurrencyEngine`, since this applies to the entire engine, whether it's running or not.

swansontec (changes requested):
This looks pretty good, but just a few small cleanups would make it better.

src/types/types.ts
```diff
@@ -893,6 +893,7 @@ export interface EdgeCurrencyEngineCallbacks {
 
 export interface EdgeCurrencyEngineOptions {
   callbacks: EdgeCurrencyEngineCallbacks
+  lightMode?: boolean // Flag indicating that the engine only needs to sync the balance and be able to spend
```
swansontec, line 896: Can we do it JavaDoc style:

```ts
/** True if we only need a balance and the ability to spend it. */
lightMode?: boolean
```

swansontec (approved):
One small comment change left to do.

PR #605, opened 2024-07-25
Add signBytes

Jon-edge:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207892031287274

### DIFF
--- a/src/core/currency/wallet/currency-wallet-api.ts
+++ b/src/core/currency/wallet/currency-wallet-api.ts
@@ -1,5 +1,6 @@
 import { div, eq, mul } from 'biggystring'
 import { Disklet } from 'disklet'
+import { base64 } from 'rfc4648'
 import { bridgifyObject, onMethod, watchMethod } from 'yaob'
 
 import {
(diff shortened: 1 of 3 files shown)


src/types/types.ts
```diff
@@ -865,6 +865,10 @@ export interface EdgeSignMessageOptions {
   otherParams?: JsonObject
 }
 
+export interface EdgeSignBytesOptions {
+  otherParams?: JsonObject
+}
```
swansontec, line 870: This type is unused. In the type definitions, you use `EdgeSignMessageOptions`. So maybe we can just delete it.

src/types/types.ts
```diff
@@ -1221,6 +1231,11 @@ export interface EdgeCurrencyWallet {
     opts?: EdgeSignMessageOptions
   ) => Promise<string>
 
+  readonly signBytes: (
```
swansontec, line 1234: Please add an `@deprecated` comment to the old `signMessage`, and stick it at the bottom of the interface. Same for the engine.

src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -532,6 +533,17 @@ export function makeCurrencyWalletApi(
       )
     },
 
+    async signBytes(
+      bytes: Uint8Array,
+      opts: EdgeSignBytesOptions = {}
+    ): Promise<string> {
+      if (engine.signBytes == null) {
+        throw new Error(`${pluginId} doesn't support signing bytes`)
```
swansontec, line 541: We talked about adding a bit in here, `if (pluginId == 'ethereum') return engine.signMessage(base16.stringify(bytes),...)`.

This would be an optional change, just to get quick support for this new feature without engine changes.

swansontec (changes requested):
One optional change, and then add `@deprecated` comments.

swansontec (approved):

PR #607, opened 2024-07-26
Randomize polling

swansontec:
This staggers the polling post-login. However, we still sync every wallet immediately at start-up time.

We might want a follow-up task to skip / delay the initial sync as well, which was not in the task description. It's likely that the "Investigate mad disklet use on login" task is actually related to the initial sync.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207731582557648

### DIFF
--- a/src/core/storage/storage-actions.ts
+++ b/src/core/storage/storage-actions.ts
@@ -11,6 +11,8 @@ import {
 } from './repo'
 import { StorageWalletStatus } from './storage-reducer'
 
+export const SYNC_INTERVAL = 30 * 1000
+
 export async function addStorageWallet(
   ai: ApiInput,
   walletInfo: EdgeWalletInfo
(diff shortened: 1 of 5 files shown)


src/core/storage/storage-actions.ts
```diff
@@ -11,6 +11,8 @@ import {
 } from './repo'
 import { StorageWalletStatus } from './storage-reducer'
 
+export const syncInterval = 30 * 1000
```
samholmes, line 14: Shouldn't this be SYNC_INTERVAL by convention?

samholmes (approved):

src/core/storage/storage-actions.ts
swansontec: Fixed.

PR #608, opened 2024-08-02
Add getMasterPublicKeys to EdgeCurrencyTools

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207863550597252

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -1031,6 +1031,9 @@ export interface EdgeCurrencyTools {
   readonly getDisplayPublicKey?: (
     publicWalletInfo: EdgeWalletInfo
   ) => Promise<string>
+  readonly getDisplayPublicKeys?: (publicWalletInfo: EdgeWalletInfo) => {
+    [key: string]: string
+  }
   readonly getSplittableTypes?: (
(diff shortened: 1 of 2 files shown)


src/types/types.ts
```diff
@@ -1031,6 +1031,7 @@ export interface EdgeCurrencyTools {
   readonly getDisplayPublicKey?: (
     publicWalletInfo: EdgeWalletInfo
   ) => Promise<string>
+  readonly getMasterPublicKeys?: () => { [key: string]: string }
```
swansontec, line 1034: You need to pass the private / public keys in as parameters. The tools have no state.

Since this is meant for the user to view, not for the app to read, `getDisplayPublicKeys` might be a better name?

The `derivePublicKey` gives the machine-readable format, and the `getDisplayPublicKey` returns a single string, with the downside that this isn't broken down and copy-pasteable. So we want a name that fits with the scheme.

swansontec (changes requested):
We need to pass in the machine-readable public keys, or it won't work.

swansontec (approved):

PR #614, opened 2024-09-02
Fetch challenge

swansontec:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207873206289034

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -1840,6 +1840,12 @@ export interface EdgeContext {
   readonly forgetAccount: (rootLoginId: string) => Promise<void>
 
   // Account creation:
+  /** Preemptively requests a CAPTCHA for account creation. */
+  readonly fetchChallenge: () => Promise<{
+    challengeId: string
+    /** If this is missing, the challenge is already solved. */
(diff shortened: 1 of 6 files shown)


src/types/types.ts
```diff
@@ -1840,6 +1840,12 @@ export interface EdgeContext {
   readonly forgetAccount: (rootLoginId: string) => Promise<void>
 
   // Account creation:
+  /** Preemptively requests a CAPTCHA for account creation. */
+  readonly fetchChallenge: () => Promise<{
+    challengeId: string
+    /** If this is missing, the challenge is already solved. */
```
Jon-edge, line 1846: Maybe an explicit status would make this more clear without having to read prop documentation

Jon-edge (approved):

src/types/types.ts
swansontec: I had it this way originally, but changed it this way for type-safety. You literally can't show the challenge to a user if there is no URL - what would go in the modal?

PR #618, opened 2024-09-10
Allow known transaction types past spam filter

peachbits:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1207973610789798

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
+- changed: Allow transactions tagged with actions through spam filter
+
 ## 2.17.0 (2024-09-11)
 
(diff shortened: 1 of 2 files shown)


swansontec (approved):
I had a sidebar with Sam about this, and we agree that this new logic is more correct. With this change, we will report transactions to the GUI if the engine reports them to us. We no longer care about 0 amounts or any other funny business - if the engine doesn't want a transaction to appear, it shouldn't report it to the core. For token transactions, this includes providing a non-null `parentNetworkFee`, which will cause the transaction to appear on the parent chain in all cases.

This will probably create a follow-on bug where 0x gasless swaps produce 0-amount, 0-fee transactions on the parent chain. It's up to the engine to fix this, not the core.

swansontec (approved):

PR #623, opened 2024-10-24
Log unknown swap errors

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208593048124847

### DIFF
--- a/src/core/swap/swap-api.ts
+++ b/src/core/swap/swap-api.ts
@@ -76,6 +76,22 @@ export async function fetchSwapQuotes(
           error => {
             pendingIds.delete(pluginId)
             log.warn(`${pluginId} gave swap error: ${String(error)}`)
+            // Log unknown errors:
+            if (isUnknownSwapError(error)) {
+              log.crash(`Unknown swap error: ${String(error)}`, {
+                error,
+                swapPluginId: pluginId,
+                request: {
+                  // Stringify to include "null"
+                  fromToken: String(request.fromTokenId),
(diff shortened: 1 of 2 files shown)


src/core/swap/swap-api.ts
```diff
@@ -232,3 +249,14 @@ function rankError(error: unknown): number {
...
+function isUnknownSwapError(error: unknown): boolean {
+  return (
+    asMaybeInsufficientFundsError(error) == null &&
+    asMaybePendingFundsError(error) == null &&
+    asMaybeSwapAboveLimitError(error) == null &&
+    asMaybeSwapBelowLimitError(error) == null &&
+    asMaybeSwapPermissionError(error) == null &&
+    asMaybeSwapCurrencyError(error) == null
+  )
+}
```
swansontec, line 262: Instead of building this whole new function, could we use `rankError(error) <= 1`?

samholmes: Oh, hmm.. The functions are semantically different and their is motivation for them to diverge in the future as we filter out noisy errors. Shall we opt to keep them distinct for this reason?

swansontec: If we give the error a score, then we know what it is, and by definition the error cannot be "unknown". Let's leverage this fact, but still keep the functions distinct:

```ts
function isUnknownSwapError(error: unknown): boolean {
  isKnownError = rankError(error) > 1
  return !isKnownError
}
```

PR #626, opened 2024-11-16
Add `getAddresses` engine method

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208512933027914

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -692,6 +692,13 @@ export interface EdgeDataDump {
   }
 }
 
+export interface EdgeAddress {
+  addressType: string // 'publicAddress' | 'segwitAddress' | 'legacyAddress' | 'fooAddress'
+  publicAddress: string
+  nativeBalance?: string
+}
+
+/** @deprecated */
 export interface EdgeFreshAddress {
   publicAddress: string
   segwitAddress?: string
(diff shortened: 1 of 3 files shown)


src/types/types.ts
```diff
@@ -693,7 +693,16 @@ export interface EdgeDataDump {
...
 }
 
+export interface EdgeAddress {
+  addressType: string // 'publicAddress' | 'segwitAddress' | 'legacyAddress' | 'fooAddress'
+  publicAddress: string
+  nativeBalance?: string
+}
+
 export interface EdgeFreshAddress {
+  allAddresses?: EdgeAddress[]
```
swansontec, line 703: So we need to do the reverse translation:

In the `EdgeCurrencyWallet.getFreshAddress` method, we should call `EdgeCurrencyEngine.getAddresses` if it exists, and then down-convert into this older `EdgeFreshAddress` format. That way, engines can stop implementing `getReceiveAddress`, but we can still provide the same core API to the GUI.

swansontec (changes requested):

src/types/types.ts
```diff
@@ -699,10 +699,8 @@ export interface EdgeAddress {
   nativeBalance?: string
 }
 
+/** Deprecated */
```
swansontec, line 702: `/* @deprecated Use getAddresses instead */`

swansontec (approved):
This looks nice. We use the `/** @deprecated */` syntax to get VSCode to show the symbol with a strikeout, but that's a simple change.

PR #627, opened 2024-11-20
DO NOT MERGE! Experiment stop files loading

samholmes:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none


paullinator (approved):

paullinator (changes requested):
Needs to successfully pass blank tx files up to GUI

src/core/currency/wallet/currency-wallet-callbacks.ts
```diff
@@ -348,8 +348,8 @@ export function makeCurrencyWalletCallbacks(
           files[txidHash],
           tx.tokenId
         )
-        if (isNew) created.push(combinedTx)
-        else if (files[txidHash] != null) changed.push(combinedTx)
+        // if (isNew) created.push(combinedTx)
+        // else if (files[txidHash] != null) changed.push(combinedTx)
```
swansontec, line 352: You probably want to do `changed.push(combinedTx)` for every transaction, so the GUI still gets updates.

The idea here is to switch between `onTxChanged` / `onNewTx`, but since we don't have files, we just want everything to go in the "existing" bucket. The `if (files[txidHash] != null)` check was there to prevent crashes, but that's no longer needed thanks to your other changes.

swansontec (approved):
This will probably work, but only "accidentally". This avoids doing an `ls` on the directory, but it doesn't actually disable loading the files themselves. That happens in `loadTxFiles`. Fortunately, without that initial `ls`, `loadTxFiles` just loops around uselessly and doesn't do anything. So, you've avoided most of the disk accesses, which means this would probably be OK for experimenting with.

To really be sure this works, I would go for the win and actually remove the call to `loadTxFiles`, which happens in `$internalStreamTransactions`. Then you know for *sure* we aren't loading any tx files from disk.

PR #628, opened 2024-11-21
Add networkFees to EdgeTransaction type

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208594536900344


.mocharc.json
swansontec, line 1: Now that we have this, could we simplify `"test": "mocha -r sucrase/register 'test/**/*.test.ts'"` to just `"test": "mocha"`.

src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -693,7 +693,9 @@ export function combineTxWithFile(
     networkFee: tx.networkFee.get(tokenId) ?? '0',
     otherParams: { ...tx.otherParams },
     ourReceiveAddresses: tx.ourReceiveAddresses,
-    parentNetworkFee: tx.networkFee.get(null) ?? '0',
+    ...(walletCurrency === currencyCode
+      ? {}
+      : { parentNetworkFee: tx.networkFee.get(null) ?? '0' }),
```
swansontec, line 704: I would prefer `parentNetworkFee: walletCurrency === currencyCode ? undefined : tx.networkFee.get(null) ?? '0'`.

There is no need to remove the field itself - setting it to `undefined` is simpler.

src/types/types.ts
```diff
@@ -526,6 +526,10 @@ export interface EdgeNetworkFee2 {
   readonly currencyPluginId: string
   readonly tokenId: EdgeTokenId
 }
+export interface EdgeNetworkFee3 {
+  readonly nativeAmount: string
+  readonly tokenId: EdgeTokenId
+}
```
swansontec, line 532: We need to get these structures sorted out, possibly requiring a separate commit / PR to clean up the mess. Let's define this:

```ts
export interface EdgeFeeAmount {
  nativeAmount: string
  tokenId: EdgeTokenId
}
```

Then we can fix:

- `EdgeTransaction.networkFees` can just be `EdgeFeeAmount[]`
- Change `EdgeSwapQuote` to use `EdgeFeeAmount & { /** @deprecated */  currencyCode: string }` instead of the old `EdgeNetworkFee`. We'll need some translation logic to upgrade / insert the `tokenId`.
- `EdgeActivationQuote` can use `EdgeFeeAmount & { /** @deprecated */ currencyPluginId: string }`. We already know the source wallet, so the `currencyPluginId` was never really needed.
- Add a `/** @deprecated use EdgeFeeAmount instead */` to both `EdgeNetworkFee` and `EdgeNetworkFee2`.

swansontec (changes requested):
This is pretty good overall, but there are a few cleanups we should take care of.

src/types/type-helpers.ts
```diff
@@ -29,3 +34,10 @@ export function upgradeCurrencyCode(opts: {
 
   return { currencyCode, tokenId: tokenId ?? null }
 }
+
+export function upgradeSwapQuote(quote: EdgeSwapQuote): EdgeSwapQuote {
+  if (quote.networkFee != null && quote.networkFee.tokenId == null) {
+    quote.networkFee.tokenId = quote.request.fromTokenId
+  }
```
swansontec, line 42: I would have put this code directly into `fetchSwapQuotes`, rather than creating a separate method, since that's less cognitive load. Optional (but recommended) change, since this way works too.

swansontec (approved):
One optional change, but this looks good now.

PR #630, opened 2024-12-10
Add `avoidCors` API to `EdgeFetchFunction`

samholmes:
This is to be able to to access `nativeFetch` in React Native
environements. This `fetchCorsForced` method will also forcibly use the
CORS servers in browser environemnts. It's idompodent in Node
environements.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208940233901583

### DIFF
--- a/src/io/browser/browser-io.ts
+++ b/src/io/browser/browser-io.ts
@@ -39,13 +39,19 @@ export function makeBrowserIo(): EdgeIo {
     }),
 
     // Networking:
-    fetch(uri: string, opts?: EdgeFetchOptions): Promise<EdgeFetchResponse> {
-      return window.fetch(uri, opts)
-    },
-    async fetchCors(
+    async fetch(
       uri: string,
       opts?: EdgeFetchOptions
     ): Promise<EdgeFetchResponse> {
+      const { corsBypass = 'auto' } = opts ?? {}
+
+      if (corsBypass === 'always') {
+        return await fetchCorsProxy(uri, opts)
+      }
+      if (corsBypass === 'never') {
+        return await window.fetch(uri, opts)
+      }
+
       const { hostname } = new URL(uri)
       const corsFailureCount = hostnameCorsProxyBlacklist.get(hostname) ?? 0
 
@@ -90,6 +96,13 @@ export function makeBrowserIo(): EdgeIo {
       // Throw the error from the first fetch instead of the one from
       // proxy server.
       throw errorToThrow
+    },
+
+    async fetchCors(
+      uri: string,
+      opts?: EdgeFetchOptions
+    ): Promise<EdgeFetchResponse> {
+      return await this.fetch(uri, opts)
     }
   }
 }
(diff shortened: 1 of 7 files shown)


src/io/browser/browser-io.ts
```diff
@@ -39,13 +39,19 @@ export function makeBrowserIo(): EdgeIo {
...
       opts?: EdgeFetchOptions
     ): Promise<EdgeFetchResponse> {
+      const { avoidCors = 'auto' } = opts ?? {}
+
+      if (avoidCors === 'never') {
+        return await fetchCorsProxy(uri, opts)
+      }
+      if (avoidCors === 'always') {
+        return await window.fetch(uri, opts)
+      }
```
swansontec, line 53: I think these two conditions are backwards.

src/io/react-native/react-native-worker.ts
```diff
@@ -172,13 +172,15 @@ async function makeIo(): Promise<EdgeIo> {
...
-      return await window.fetch(uri, opts)
-    },
+      const { avoidCors = 'auto' } = opts ?? {}
+
+      if (avoidCors === 'never') {
+        return await nativeFetch(uri, opts)
+      }
+      if (avoidCors === 'always') {
+        return await window.fetch(uri, opts)
+      }
```
swansontec, line 182: I think these are backwards too. Saying "never avoid CORS" means using the browser, and saying "always avoid CORS" means using the native fetch function.

src/types/types.ts
```diff
@@ -76,6 +76,8 @@ export interface EdgeIo {
   /**
    * This is like `fetch`, but will try to avoid CORS limitations
    * on platforms where that may be a problem.
+   *
+   * @deprecated - Use `fetch` instead with `avoidCors: 'always'` option.
```
swansontec, line 80: Actually `avoidCors: 'auto'` would be the correct equivalent. We could actually simplify the message to `@deprecated Use EdgeIo.fetch instead, which now includes CORS avoidance by default`

swansontec (changes requested):
In the commit message, you say "idempotent" when you really mean "equivalent". The word "idempotent" means once you've done something once, doing it more times has no effect, like applying the `Math.abs` function.

Also, please edit the `loginFetch` function to use `avoidCors: 'never'`.

Also, in root.ts, find the `makeSyncClient` call and do `fetch: (url, opts) => io.fetch(url, {...opts, avoidCors: 'never'})`, to keep the behavior the same.

src/io/browser/browser-io.ts
samholmes: If we are "never avoiding cross-origin request sharing" then we are "allowing cross-origin request sharing" which means we are allowing sharing between our origin (null) and the server's, so we should not let CORS policies block.

If we're are "always avoiding cross-origin request sharing" then we are "never cross-origin request sharing" which means if the server doesn't share our origin (null) then it should fail.

This is the logic I understood it as. Which is why it seemed a bit confusing to read and may not be obvious to the user if you're deriving an opposite conclusion from what I did when reading the API.

swansontec: Went with the "bypass" name instead, to make it clear what is meant.

swansontec (approved):

PR #635, opened 2025-01-10
Add chainDisplayName and assetDisplayName

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1208490257069911
  - https://app.asana.com/0/0/1208490257069916

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -454,7 +454,12 @@ export type EdgeObjectTemplate = Array<
 export interface EdgeCurrencyInfo {
   // Basic currency information:
   readonly pluginId: string
+
+  /** deprecated: refers to chain, use chainDisplayName instead */
   displayName: string
+
+  chainDisplayName: string
+  assetDisplayName: string
   walletType: string
 
   // Native token information:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 ## Unreleased
 
+- added: `EdgeCurrencyInfo.chainDisplayName` and `EdgeCurrencyInfo.assetDisplayName`
+- deprecated: `EdgeCurrencyInfo.displayName`
+
 ## 2.22.2 (2025-01-03)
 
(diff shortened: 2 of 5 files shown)


src/core/currency/wallet/currency-wallet-api.ts
```diff
@@ -110,6 +110,12 @@ export function makeCurrencyWalletApi(
       otherMethods[name] = (...args) => method(walletInfo.keys, ...args)
     }
   }
+  if (plugin.currencyInfo.chainDisplayName == null) {
+    plugin.currencyInfo.chainDisplayName = plugin.currencyInfo.displayName
+  }
+  if (plugin.currencyInfo.assetDisplayName == null) {
+    plugin.currencyInfo.assetDisplayName = plugin.currencyInfo.displayName
+  }
```
swansontec, line 118: This upgrade logic won't run until we have a wallet, which is too late, because the `currencyInfo` is also available on the `EdgeCurrencyConfig` object as well.

To do the upgrade as early as possible, please more this logic to the `CORE_PLUGINS_ADDED` redux action in plugins-reducer.ts, which is where the core first takes the wrapping paper off the plugins.

swansontec (changes requested):
We need to fix the upgrade logic.

src/types/types.ts
```diff
@@ -454,7 +454,12 @@ export type EdgeObjectTemplate = Array<
 export interface EdgeCurrencyInfo {
   // Basic currency information:
   readonly pluginId: string
+
+  /** deprecated: refers to chain, use chainDisplayName instead */
   displayName: string
+
+  chainDisplayName?: string
+  assetDisplayName?: string
```
swansontec, line 462: Please make these two mandatory, not optional. That way the GUI can use them reliably. Yes, it's maybe a bit annoying for the plugins on the next core upgrade, but we consider this a non-breaking change according to our plugin compatibility policy, since things still work at run-time.

swansontec (changes requested):

swansontec (approved):

PR #643, opened 2025-03-14
Integrate Change Server

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1209129370445709


src/core/currency/wallet/currency-wallet-pixie.ts
```diff
@@ -236,7 +234,46 @@ export const walletPixie: TamePixie<CurrencyWalletProps> = combinePixies({
...
+  syncNetwork: filterPixie(
+    (input: CurrencyWalletInput) => {
+      let syncNetworkTask: PeriodicTask
+
+      return {
+        update() {
+          const { currencyInfo, walletInfo, publicWalletInfo } =
+            input.props.walletState
+          const { engine } = input.props.walletOutput
+          if (engine != null) {
```
swansontec: Before, we waited for `startEngine` to resolve before we started the `syncNetwork` task. Now we *might* call `syncNetwork` before `syncNetwork` has finished (it's racey).

We could fix the race by having `engineStarted` call `input.props.onOutput(boolean)` to indicate that the `startEngine` has finished, and we're ready to sync. Then have the new `syncNetwork` pixie wait for `input.props.walletOutput.engineStarted` to be true.

src/util/json-rpc.ts
swansontec, line 1: This file looks out-of-sync with the copy in edge-change-server.

src/core/currency/wallet/currency-wallet-reducer.ts
```diff
@@ -178,6 +192,22 @@ const currencyWalletInner = buildReducer<
     }
   ),
 
+  changeServiceSubscriptions(state = [], action) {
+    if (action.type === 'CURRENCY_ENGINE_UPDATE_CHANGE_SERVICE_SUBSCRIPTIONS') {
+      console.log('CURRENCY_ENGINE_UPDATE_CHANGE_SERVICE_SUBSCRIPTIONS', [
+        ...state,
+        ...action.payload.subscriptions
+      ])
```
swansontec, line 200: Leftover logging.

src/core/currency/wallet/currency-wallet-callbacks.ts
```diff
@@ -287,6 +287,25 @@ export function makeCurrencyWalletCallbacks(
...
+          address,
+          status: 'subscribing' as const
+        }))
+      ]
+      input.props.dispatch({
+        type: 'CURRENCY_ENGINE_UPDATE_CHANGE_SERVICE_SUBSCRIPTIONS',
+        payload: {
+          subscriptions,
+          walletId
+        }
```
swansontec, line 305: There is no way for the engine to remove an address. Anything passed in here gets added to the list we already have. This could be troublesome for chains like Hedera where activating the wallet causes its address to permanently change.

We could solve this by adding a boolean flag to this redux action - setting `isFromEngine` to `true` would cause the reducer to stomp away anything the engine didn't mention.

However, I don't think we need to worry about this too much - it's not going to be relevant for ETH or BTC, which either don't change or only add addresses. Even with a chain like Hedera, it should be harmless extra noise to have a few junk subscriptions floating around. So let's not fix it, but just mark it with a comment and move on. We can come up with a proper fix once we have a chain that cares, and we know more about its requirements.

src/core/currency/change-server-protocol.ts
```diff
@@ -0,0 +1,69 @@
...
+  clientMethods: {
+    update: {
+      asParams: asSubscribeParams
+    },
+    pluginConnect: {
+      asParams: asObject({ pluginId: asString })
+    },
+    pluginDisconnect: {
+      asParams: asObject({ pluginId: asString })
+    }
```
swansontec, line 67: This is the old protocol, so we'll have to update to `subLost` and then re-review.

swansontec (changes requested):
This looks really good. We need to update to the final change-server protocol, but what we have so far is clean & nice.

src/core/currency/change-server-protocol.ts
samholmes: Added the fixup commit

src/util/json-rpc.ts
samholmes: Should be up-to-date now.

src/core/currency/wallet/currency-wallet-pixie.ts
```diff
@@ -233,6 +234,10 @@ export const walletPixie: TamePixie<CurrencyWalletProps> = combinePixies({
             // Turn synchronous errors into promise rejections:
             startupPromise = Promise.resolve()
               .then(() => engine.startEngine())
+              .then(() => {
+                // Signal that engine startup is complete:
+                input.onOutput({ ...walletOutput, engineStarted: true })
+              })
```
swansontec, line 240: I'm pretty sure `combinePixies` takes care of the merging, and we only need to do `input.onOutput(true)` here.

Since you've put this extra `{}` in here, I'm guessing your debugger will show something like:

```js
{
  walletApi: {...},
  engine: {...},
  engineStarted: {
    walletApi: {...},
    engine: {...},
    engineStarted: true
  }
}
```

Can we verify that it's messed up in this way, and then fix it?

src/core/currency/wallet/currency-wallet-pixie.ts
```diff
@@ -249,12 +254,13 @@ export const walletPixie: TamePixie<CurrencyWalletProps> = combinePixies({
             startupPromise
               .then(() => engine.killEngine())
               .catch(error => input.props.onError(error))
-              .then(() =>
+              .then(() => {
+                input.onOutput({ ...walletOutput, engineStarted: false })
```
swansontec, line 258: This would also need to be fixed if my previous comment turns out to be correct.

src/util/json-rpc.ts
swansontec: Are you sure? I still don't see the blockbook subscription hack.

swansontec (commented):
I found some stuff to check before I approve.

src/core/currency/wallet/currency-wallet-pixie.ts
samholmes: Good catch.

swansontec (approved):

PR #654, opened 2025-05-12
Fix accidental duress stash creation in loginWithKey

samholmes:
The `loginWithKey` method has a bug where it will call `ensureAccountExists` for the duress account if `duressMode` option is set to true. If we don't have a duress account for the account being logged into, then we shouldn't be creating an duress account from within `loginWithKey`.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210230037447340

### DIFF
--- a/test/core/login/login.test.ts
+++ b/test/core/login/login.test.ts
@@ -547,6 +547,28 @@ describe('duress', function () {
     expect(topicAccount2.isDuressAccount).equals(true)
   })
 
+  it('Avoid creating duress account when using loginWithKey', async function () {
+    const world = await makeFakeEdgeWorld([fakeUser], quiet)
+    const context = await world.makeEdgeContext(contextOptions)
+    const otherAccount = await context.createAccount({
+      username: 'new-user',
+      pin: '1111'
+    })
+    const loginKey = await otherAccount.getLoginKey()
+    await otherAccount.logout()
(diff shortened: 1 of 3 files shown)


swansontec (approved):

test/core/login/login.test.ts
```diff
@@ -523,6 +523,50 @@ describe('duress', function () {
...
+    await otherAccount.logout()
+    // Activate duress mode:
+    const duressAccount = await context.loginWithPIN('new-user', '0000')
+    await duressAccount.logout()
+
+    // Login to the main account using password:
+    const topicAccount = await context.loginWithPassword(
+      fakeUser.username,
+      fakeUser.password
+    )
```
swansontec, line 544: Also test login with key?

swansontec (approved):

test/core/login/login.test.ts
samholmes: I did this in the first commit: https://github.com/EdgeApp/edge-core-js/pull/654/commits/df9e25b6e36ae1d3a872907d0573a754da11ba80#diff-1fe6c58a4991bd093efe5acca148ea511af9af714e0e17ff1ae54e9865f45fccR550

The second commit added the other test cases with other login methods.

PR #657, opened 2025-05-15
Extend onSubscribeAddresses to include checkpoints

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210267237282368

### DIFF
--- a/src/core/currency/wallet/currency-wallet-callbacks.ts
+++ b/src/core/currency/wallet/currency-wallet-callbacks.ts
@@ -291,14 +291,27 @@ export function makeCurrencyWalletCallbacks(
       })
     },
 
-    onSubscribeAddresses(addresses: string[]) {
+    onSubscribeAddresses(
+      paramsOrAddresses:
+        | Array<{ address: string; checkpoint?: string }>
+        | string[]
+    ) {
+      const params = paramsOrAddresses.map(param =>
+        typeof param === 'string'
+          ? {
+              address: param,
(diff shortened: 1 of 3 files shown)


src/core/currency/wallet/currency-wallet-callbacks.ts
```diff
@@ -291,14 +291,27 @@ export function makeCurrencyWalletCallbacks(
...
+    onSubscribeAddresses(
+      paramsOrAddresses:
+        | Array<{ address: string; checkpoint?: string }>
+        | string[]
+    ) {
+      const params = paramsOrAddresses.map(param =>
+        typeof param === 'string'
+          ? {
+              address: param,
+              checkpoint: 'latest'
```
swansontec, line 303: I would use `undefined` instead of `'latest'`. That is the existing pattern in the change server - the checkpoint is optional, and skipping it means we either don't know or care, so check everything.

In either case, "latest" is the opposite of what we want. Even if we did a string, it should be "oldest", to indicate that we don't know the start point, so go back to the beginning.

But yeah, trying to do `try { parseInt }` is a little annoying, so let's keep the string contents 100% under plugin control. We don't want to insert our own strings here.

swansontec (changes requested):
One change needed.

swansontec (approved):

PR #658, opened 2025-05-19
Add `getTokenDetails` to currency tools

peachbits:
### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none


src/core/account/custom-tokens.ts
```diff
@@ -32,6 +33,21 @@ const asMaybeContractLocation = asMaybe(
...
+  pluginId: string,
+  filter: EdgeGetTokenDetails
+): Promise<EdgeToken[]> {
+  // The normal code path:
+  const tools = await getCurrencyTools(ai, pluginId)
+  if (tools.getTokenDetails == null) return []
+
+  return await tools.getTokenDetails(filter)
+}
+
```
swansontec, line 50: This is simple enough, I would just put it inline in plugin-api.ts. There is no real logic in this function, just a wrapper around a tools method.

swansontec (approved):
One small cleanup (maybe)

src/types/types.ts
```diff
@@ -426,6 +426,14 @@ export interface EdgeTokenInfo {
   multiplier: string
 }
 
+/**
+ * Search the blockchain for a token.
+ */
+export interface EdgeGetTokenDetails {
```
samholmes, line 432: Also, shouldn't this be named `EdgeGetTokenDetailsParams` or `EdgeGetTokenDetailsFilter`?

samholmes (dismissed):

PR #659, opened 2025-05-20
Sam/rest o duress

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none


src/core/context/context-api.ts
```diff
@@ -191,6 +191,8 @@ export function makeContextApi(ai: ApiInput): EdgeContext {
       // Attempt to log into duress account if duress mode is enabled:
       if (ai.props.state.clientInfo.duressEnabled) {
         const duressAppId = appId + '.duress'
+        const stash =
```
swansontec, line 194: We already called `getStashByUsername` above, so it could be `stash ?= getEmptyStash(username)` on this line.

src/core/login/password.ts
```diff
@@ -89,6 +90,11 @@ async function loginPasswordOnline(
       return decrypt(passwordBox, passwordKey)
     }
   )
+  // Disable duress mode if it is enabled:
+  if (ai.props.state.clientInfo.duressEnabled) {
+    await disableDuressMode(ai)
+  }
+  return sessionKey
```
swansontec: This makes 0 sense, but I see you do it properly in a later commit. Just drop this one, or squash it with the other one.

src/core/login/login-stash.ts
```diff
@@ -75,20 +75,42 @@ export interface LoginStash {
...
+  const temporaryOverrides: LoginStash[] = []
   const paths = await disklet.list('logins').then(justFiles)
   for (const path of paths) {
     try {
-      out.push(asLoginStash(JSON.parse(await disklet.getText(path))))
+      const stash = asLoginStash(JSON.parse(await disklet.getText(path)))
+      if (path.endsWith('-temp.json')) {
+        temporaryOverrides.push(stash)
+        continue
+      }
```
swansontec, line 96: We don't want to add a whole new file, because disk is slow. Instead, we want to add fields to the `LoginStash` type. Something like:

```ts
interface LoginStash {
  ...
  // Tell the truth if these are undefined:
  pinEnabledLie?: boolean
  duressEnabledLie?: boolean
  ...
}
```

And then where we would normally do `pinEnabled = <existing-logic>`, we just do `pinEnabled = pinEnabledLie ?? <existing-logic>`. Pretty straightforward.

swansontec (changes requested):

src/core/login/password.ts
samholmes: If a user logs into an account using username and password and the account is not on the device, then the device would need to authenticate with login server for the login stash. If the device duress mode setting was enabled in the client file, it wouldn't be disabled because the only way to disable it is by entering a correct pin. However, they'd be logging into the the account in non-duress mode, so the device is in a qusi-duress mode state where the device is in duress mode, but the account is not the duress mode account.... There are two options:

1. Implement logging into to duress stash (if present) even when logging into an account via the login server authentication (this changes the spec)
2. Disable duress mode when logging with password via the login server (no change to the spec)

In this commit, I chose to #2 because its a bug and this is a fix. I added the test to enforce the spec.

In the later commit, I change the spec. I made it a separate commit to review whether we want to change the spec. It sounds like you're for the later commit, but perhaps not. Now that I explained it thoroughly and you understand there is a spec change, perhaps you have a change of opinion?

src/core/context/context-api.ts
samholmes: We need to get the stash again in order to find the `duressStash` for the case where we don't have a stash until we call `loginPassword`. So, I think you're asking for just a type assertion, which is fine. I'll fixup.

src/core/login/login-stash.ts
samholmes: What you propose will work, but would this mean that "online logins" will stomp on these fields? If so, then my [next commit](72d334d13136412ed0dc4f7e54657ac4a0577b68) which changes the spec is not possible, which means that the [first commit](fcd0aa5189888947aa38c0489caa2a53c98aa597) must be accepted in full.

I can try you're suggestion here and if it doesn't stomp, then no further discussion needed.

swansontec: Server logins never stomp the stash. That's why we use `updateTree` and require an existing stash - we can always preserve the existing fields by copying them over as part of the update.

src/core/login/password.ts
swansontec: We need to do option 1. That is the only possible correct behavior, because Edge has another unwritten rule that offline & online password logins work identically. If offline password login can access the duress account, the online login *must* do the same thing.

This isn't a spec change, it's required behavior. If we aren't doing that, then it's just another bug that must be fixed.

swansontec (approved):

src/core/context/context-api.ts
swansontec: Ah, missed that. Carry on.

PR #661, opened 2025-05-20
Sam/duress check pin

samholmes:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none


src/core/login/pin2.ts
```diff
@@ -173,14 +173,18 @@ export async function changePin(
...
-  pin: string
+  pin: string,
+  forDuressAccount?: boolean
 ): Promise<boolean> {
-  const { appId, loginId, username } = login
+  const { loginId, username } = login
   if (username == null) return false
 
+  const appId =
+    forDuressAccount === true ? login.appId + '.duress' : login.appId
```
swansontec, line 183: If we are already in duress mode, the appId would be `.duress.duress`.

src/core/login/pin2.ts
```diff
@@ -173,14 +173,18 @@ export async function changePin(
...
+
   // Find the stash to use:
   const { stashTree } = getStashById(ai, loginId)
-  const stash = findPin2Stash(stashTree, appId)
+  const stash = findPin2StashSpecific(stashTree, appId)
```
swansontec, line 187: It seems like this would break the existing logic.

I would have expected:

```ts
const stash = forDuressAccount
  ? searchTree(stashTree, stash => stash.appId === appId + '.duress')
  : findPin2Stash(stashTree, appId)
```

Then the existing code-path stays unchanged.

Of course, we'd also want to fix the `.duress.duress` logic error mentioned earlier.

src/types/types.ts
```diff
@@ -1724,7 +1724,10 @@ export interface EdgeAccount {
 
   // Verify existing credentials:
   readonly checkPassword: (password: string) => Promise<boolean>
-  readonly checkPin: (pin: string) => Promise<boolean>
+  readonly checkPin: (
+    pin: string,
+    opts?: { forDuressAccount?: boolean }
```
swansontec, line 1729: I generally like to avoid the words `login` or `account` in core parameter names. Just `forDuress` would also be clear.

swansontec (changes requested):

src/core/login/pin2.ts
samholmes: Ah, actually `login` is always `loginTree` (a root). If it's not a root, then `if (username == null) return false` would prevent this line from being reached. Unless I misunderstand that non-roots can have a username field? If so, then it's potentially non-root, and `.duress.duress` is possible, but then the stash search would fail, and we'd get an exception out of this function. So generally this is safe.

You did make me think of another test case. I added it and included changes.

src/core/login/pin2.ts
samholmes: This would fail because if the loginStash is the `.duress` account, then `searchTree` would fail to find it in that case. `findPin2StashSpecific` is invented for this reason. "Specific" here is meant to mean target "appId" specifically passed; `findPin2Stash` doesn't return the stash specifically for appId.

src/types/types.ts
samholmes: Oh shoot. I agree, but we already use `forDuressAccount` in or API. We either:

1. Are inconsistent and use `forDuress` here.
2. Are consistent and use `forDuressAccount` here
3. Are consistent, use `forDuress` here, and add it everywhere else with `forDuressAccount` deprecated.

src/core/login/pin2.ts
samholmes:
```
const stash = forDuressAccount ? findPin2StashSpecific(stashTree, appId) : findPin2Stash(stashTree, appId)
```

src/types/types.ts
samholmes: We're going with #2

PR #673, opened 2025-07-23
Added `allowChainedPending` flag to `EdgeSpendInfo`

Jon-edge:
To allow spends with chained nonces despite existing pending transactions.

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

 none

### Description

 none

---
  - https://app.asana.com/0/0/1210803607950460

### DIFF
--- a/src/types/types.ts
+++ b/src/types/types.ts
@@ -702,6 +702,8 @@ export interface EdgeSpendInfo {
   /** @deprecated Use EdgeCurrencyWallet.accelerate instead */
   rbfTxid?: string
   skipChecks?: boolean
+  /** Allow creating transactions with chained nonces even if pending transactions exist */
+  allowChainedPending?: boolean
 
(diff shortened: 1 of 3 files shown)


src/types/types.ts
```diff
@@ -702,6 +702,8 @@ export interface EdgeSpendInfo {
   /** @deprecated Use EdgeCurrencyWallet.accelerate instead */
   rbfTxid?: string
   skipChecks?: boolean
+  /** Allow creating transactions with chained nonces even if pending transactions exist */
+  allowChainedPending?: boolean
```
swansontec, line 706: This is not what we discussed. Since this is a 100% EVM problem, so it shouldn't be in the core types. It should be in the `otherParams`. So this PR needs to be dropped.

swansontec (changes requested):

PR #675, opened 2025-08-01
Fix transpilation bug

samholmes:
Targeting ES5 transpilation causes a bug with looping over iterables. The transpilation assumes all loops are over arrays, and so it loops using `.length` properties. We cannot loop over iterables, therefore we must convert the iterable to an array before looping.

This fixes a major bug with syncing caused by transpilation of the source.

### CHANGELOG

- [ ] Yes
- [ ] No

### Dependencies

 none

### Description

 none

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## Unreleased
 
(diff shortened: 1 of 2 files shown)


swansontec (approved):
Yikes! Great catch!

PR #712, opened 2026-03-03
Remove NYM mixFetch per-host queue

bobbythelobster:
## Summary
- remove custom per-host request serialization in 
- keep  export as a thin pass-through to 
- remove stale comments at NYM callsites in browser/react-native IO
- add changelog note under Unreleased

## Why
 was upgraded in 2.43.1 for native concurrency improvements. Keeping our own host queue can unnecessarily serialize NYM traffic and increase timeout risk.

## Validation
- 
- 

## Risk
- behavior changes from serialized per-host to library-managed concurrency; relies on current mix-fetch behavior.

### DIFF
--- a/src/util/nym.ts
+++ b/src/util/nym.ts
@@ -17,24 +17,6 @@ export const mixFetchOptions: SetupMixFetchOps = {
 // MixFetch initialization state
 let mixFetchInitPromise: Promise<IMixFetch> | null = null
 
-// Per-host request queue to handle mixFetch's one-request-per-host limitation
-// Maps host -> Promise that resolves when current request completes (for chaining)
-const hostRequestChains = new Map<string, Promise<Response>>()
-
-/**
- * Extract the host:port from a URI for queue keying
- */
-function getHostKey(uri: string): string {
-  try {
-    const url = new URL(uri)
-    const port =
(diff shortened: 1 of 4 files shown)


src/util/nym.ts
```diff
@@ -58,31 +40,11 @@ export async function initMixFetch(log: EdgeLog): Promise<IMixFetch> {
 }
 
 /**
- * Queue-wrapped mixFetch that serializes requests per host.
- * mixFetch only allows one concurrent request per host, so we chain them.
+ * Thin wrapper around mixFetch with pre-configured options.
  */
 export async function queueMixFetch(
```
samholmes: We don't need queueMixFetch at all. We can just inline mixFetch where queueMixFetch is being invoked.

samholmes (changes requested):

src/util/nym.ts
j0ntz: Addressed on the current branch. `queueMixFetch` has been removed (commit cb91393c) and `mixFetch` is now called directly where the queue was previously invoked: src/io/browser/browser-io.ts:55 and src/io/react-native/react-native-worker.ts:182 (commit d8b74e0e). `src/util/nym.ts` no longer exports a queue wrapper, only `initMixFetch`/`mixFetchOptions`. Leaving the thread unresolved for the PR author to close.

PR #736, opened 2026-08-14
Send attestation tokens on login-server requests

paullinator:
## Summary
- Add `EdgeContext.setAttestationToken` and attach `x-attestation-token` on login-server `/api/` requests so attested clients can receive CAPTCHA rate overrides.
- Treat empty-string tokens as cleared (header omitted).
- Allow private LAN IPv4 hosts for `http`/`ws` server overrides (still reject them for `https`/`wss`).

## Test plan
- [ ] `attestation-header` tests: set / clear / empty-string token behavior
- [ ] `validateServer` tests: LAN `http` accepted; public IPs and `https` LAN rejected
- [ ] Login-server requests include `x-attestation-token` only when a token is set

---

> [!NOTE]
> **Medium Risk**
> Touches login-server request headers and server URL validation used at context init; mistakes could weaken override rules or break local dev server configs.
> 
> **Overview**
> Adds **`EdgeContext.setAttestationToken`** so apps can supply a device attestation JWT that is sent as **`x-attestation-token`** on subsequent login-server `/api/` requests (for CAPTCHA rate overrides). Empty or cleared tokens omit the header.
> 
> **`validateServer`** now allows RFC1918 and loopback IPv4 hostnames for **`http`/`ws`** overrides only; **`https`/`wss`** still require localhost or `*.edge(test)?.app`, with tests for spoofed DNS names like `10.evil.com`.
> 
> Fake-world **`allowNetworkAccess`** no longer routes all non-Edge URLs through real `io.fetch` via `validateServer`; only **login, info, and sync** `*.edge.app` hosts stay on the in-memory fake server—change servers and private LAN use real fetch.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 552ba2184696f0798183b31f635c44407ba63b8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/util/validateServer.ts
+++ b/src/util/validateServer.ts
@@ -1,11 +1,13 @@
 /**
- * We only accept *.edge.app or localhost as valid domain names.
+ * We only accept *.edge.app, localhost, or (for http/ws only) private LAN IPv4.
+ * https/wss still require localhost or *.edge(test)?.app; private IPs are not
+ * accepted on secure schemes.
  */
 export function validateServer(server: string): void {
   const url = new URL(server)
 
   if (url.protocol === 'http:' || url.protocol === 'ws:') {
-    if (url.hostname === 'localhost') return
+    if (isPrivateHost(url.hostname)) return
   }
   if (url.protocol === 'https:' || url.protocol === 'wss:') {
     if (url.hostname === 'localhost') return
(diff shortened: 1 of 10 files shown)


src/util/validateServer.ts
```diff
@@ -1,18 +1,48 @@
...
+ * We only accept *.edge.app, localhost, or (for http/ws only) private LAN IPv4.
+ * https/wss still require localhost or *.edge(test)?.app — private IPs are not
+ * accepted on secure schemes.
  */
 export function validateServer(server: string): void {
   const url = new URL(server)
 
   if (url.protocol === 'http:' || url.protocol === 'ws:') {
-    if (url.hostname === 'localhost') return
+    if (isPrivateHost(url.hostname)) return
```
j0ntz, line 10: Side effect on fake-world routing: `makeFakeWorld` with `allowNetworkAccess: true` routes fetches by "passes validateServer -> fakeFetch, throws -> real network" (src/core/fake/fake-world.ts). Private-IP URIs now pass, so a fake-world test pointed at a real LAN dev server (the exact use case this PR enables) silently gets answered by the in-memory fake server instead of the network. If that is not intended, the fake-world path may want to keep the old localhost-only check.

src/util/validateServer.ts
```diff
@@ -1,18 +1,48 @@
 /**
- * We only accept *.edge.app or localhost as valid domain names.
+ * We only accept *.edge.app, localhost, or (for http/ws only) private LAN IPv4.
+ * https/wss still require localhost or *.edge(test)?.app — private IPs are not
```
j0ntz: Nit: em-dash in the comment; repo convention is zero em-dashes in committed code (comma or semicolon instead). Full ruleset: https://github.com/EdgeApp/edge-dev-agents/blob/main/.cursor/skills/no-slop/SKILL.md

src/util/validateServer.ts
j0ntz: Fixed in 552ba21, and in a better place than I suggested: fake-world no longer consults `validateServer` at all, so test routing stops tracking a production policy function. The explicit `isFakeAccountInfrastructure` allowlist is the right decoupling.

I checked the regex against the hosts this repo actually ships: `login1/login2`, `info1/info2`, `sync-us1` through `sync-us6`, `sync-eu` (root.ts) and the fake server's own `info-fake1`, `sync-fake1..3` all match.

Two intentional behavior changes worth naming, both consistent with the updated doc comment: under `allowNetworkAccess`, `localhost` and non-account `*.edge.app` hosts (the `cors1..4.edge.app` proxies, for instance) now reach the real network instead of the fake server.

src/util/validateServer.ts
j0ntz: Fixed in 552ba21.

j0ntz (approved):

PR #738, opened 2026-08-21
Back off NYM mixFetch setup after a failure

j0ntz:
### Technical Design Document

[zano-migration-off-module-queue.md](https://github.com/EdgeApp/react-native-zano/blob/jon/zano-migration-tdd/src/docs/zano-migration-off-module-queue.md)

### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

none

### Description

QA reported an iOS test account logging itself out every minute or so with auto-logout set to one hour. The attached logs show it is not the logout timer: each re-login is preceded by `EdgeContext closed` and then `EdgeContext opened` about a second later, with no JS bundle relaunch. That is `EdgeCoreWebView.webViewWebContentProcessDidTerminate` calling `visitPage()`, so iOS killed the core WebView's WebContent process for memory and the core reloaded itself, dropping the session.

What exhausts the memory is the NYM mixFetch setup path. `initMixFetch` sets its cached promise back to null whenever a setup fails, so the next privacy fetch immediately builds another client. `createMixFetch` spawns a web worker that loads about 13 MB of WASM before it ever contacts the gateway, and the library exposes no way to terminate that worker, so each failed attempt leaves one behind. With an engine poll loop driving a request every few seconds against a gateway that is refusing the client, roughly thirty workers accumulate inside a minute and the WebContent process is killed.

In the reported logs every crash-looping session carries `Initializing mixFetch...` about every 2.5 seconds, each ending in `gateway client error (5rXcNe2a...)`; the one long healthy session in the same file has no mixFetch lines at all.

This change gives a failed setup a cooldown, starting at 30 seconds and doubling to a 5 minute ceiling. Callers arriving during the cooldown reject immediately instead of building another client, so a dead gateway costs one worker per window instead of one per request.

The reported symptom needs both an account with NYM privacy enabled and a failing gateway, which is why it does not appear on a default account.

### What review changed

Four rounds with @peachbits reshaped the details, and the notes are worth carrying here because each one is load-bearing:

- **The cooldown is armed before the library is touched.** The failure handler is no longer `async`: it nulls the cached promise, records the error and sets `retryAfter` first, then fires `disconnectMixFetch().catch(() => {})` without awaiting it. Comlink's request path (`index.js:342-351`) builds a promise with no reject and no timeout, so a disconnect issued against a worker the OS already killed never settles. Awaiting it stranded every line below, the cooldown included, which is the freeze this PR exists to prevent. A test makes the fake disconnect hang forever and fails against the awaited form.
- **A cooling-down caller gets its own error, not the stored instance.** Re-throwing `lastError` handed callers a stack from a setup that ended minutes ago, so a 0 ms rejection read in a crash report as a fresh 60 second timeout. The rejection now names the remaining wait and carries the original as `cause`, which also drops the assumption that `lastError` is an `Error` at all: the library rejects with a raw `MessageEvent` on a worker error.
- **One breadcrumb per cooldown window, not one per refused caller.** The GUI caps Sentry at `maxBreadcrumbs: 25` (`src/app.ts:45`), so a poll loop refusing every few seconds would evict the app's whole breadcrumb history inside a minute, blinding the crash report this fix exists to inform. The breadcrumb moved to the arming site and names the window actually armed. A test asserts the cardinality, since the failure mode is silent.
- **The setup routine is built by a factory.** `makeMixFetchSetup(now = () => Date.now())` closes over the cooldown state, and `export const initMixFetch = makeMixFetchSetup()` keeps both io call sites unchanged. That removed a `resetMixFetchForTests` export and a mutable module clock from the shipped surface, and gives each test fresh state for free. Its one caveat is documented on the function: the cooldown state is per instance while the client, its worker and `window.__mixFetchGlobal` are process-global, so the single exported instance is the only supported arrangement.
- **Dropped: clearing the cooldown on success.** Those resets were unreachable. The cached promise is only nulled in the failure handler, so after a success the guard at the top of the routine is false forever and nothing reads the cooldown state again. The test that appeared to cover them went with them.

### Testing

`test/util/nym.test.ts` drives the module against a controlled clock and counts how many clients get built. Five cases: twenty-five seconds of requests after a failure build exactly one client; a cooling-down caller is rejected with the naming error and the preserved `cause`; the breadcrumb fires once per window across eleven refusals and again on the next window; a retry happens once the window expires and the wait doubles after the second failure; and the cooldown still arms when the cleanup disconnect never settles. Each case was checked against the pre-fix code it guards.

The full suite reports 168 passing, `npm run types` is clean, and `verify-repo.sh` passed.

Driven in the app on an iOS simulator, with `edge-core-js` linked into `edge-react-gui` via updot and Ethereum's Network Privacy set to Nym Mixnet. The gateway failure is forced with a temporary uncommitted edit that rejects the setup after 2s, while still building the real client so each attempt still spawns its worker. Over 3m56s the engine's poll loop produced 82 privacy fetches, and the cooldown answered them with 4 client builds and 4 breadcrumbs, at 30s, 60s, 120s and 240s. The same drive with `RETRY_BASE_MS` set to 0 produced 47 builds and 47 breadcrumbs in 1m55s, one every ~2.4s, which is the cadence the QA logs carried. Screenshots are in the test-evidence comment below; both edits are reverted.

Still not verified: that WebContent memory stays flat across a long outage on a real device. The build rate is what this change controls and it is now measured, but the resulting memory ceiling was not observed on hardware.

Asana: https://app.asana.com/0/1215088146871429/1217733586839287
Parent: https://app.asana.com/0/1215088146871429/1217559756673909

---

> [!NOTE]
> **Medium Risk**
> Changes NYM privacy fetch initialization on failure paths; mis-tuned backoff could delay recovery when the gateway returns, but the behavior only affects accounts with NYM privacy and a failing setup.
> 
> **Overview**
> Fixes an iOS symptom where NYM privacy with a **failing gateway** looked like repeated logouts: each failed **`createMixFetch`** left a WASM web worker alive, and **`initMixFetch`** cleared its cached promise on failure so poll-driven privacy fetches spawned a new worker every few seconds until the WebView process was killed for memory.
> 
> **`makeMixFetchSetup`** now owns per-instance cooldown state; production still uses **`export const initMixFetch = makeMixFetchSetup()`**. After a failed setup (including the existing 60s timeout), new attempts are **refused until a backoff window** passes—**30s initially, doubling up to 5 minutes**—instead of calling **`createMixFetch`** again. Callers during cooldown get a **fresh “cooling down” error** with the original failure on **`cause`**, not a re-thrown stale error. The cooldown is **armed before** best-effort **`disconnectMixFetch`** / global cleanup, which are **not awaited** so a hanging disconnect cannot block later callers. **One Sentry breadcrumb** is emitted per cooldown window. **`test/util/nym.test.ts`** covers client count, errors, breadcrumbs, backoff, and hang behavior with a fake mix-fetch module and injectable clock.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eec286f7556166d17f1f9f6410880f55f3cec23e. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/src/util/nym.ts
+++ b/src/util/nym.ts
@@ -32,54 +32,148 @@ export const mixFetchOptions: SetupMixFetchOps = {
  */
 const SETUP_TIMEOUT_MS = 60000
 
-// MixFetch initialization state
-let mixFetchInitPromise: Promise<IMixFetch> | null = null
+/**
+ * How long to refuse new setups after one fails, and the ceiling that wait
+ * doubles up to.
+ *
+ * Every `createMixFetch` spawns a web worker holding megabytes of WASM before
+ * it ever contacts the gateway, and the library exposes no way to terminate
+ * that worker, so a failed setup leaves one behind. Retrying on each request
+ * therefore costs memory per attempt: with a dead gateway and a poll loop
+ * driving requests every few seconds, the workers accumulate until the host
+ * kills the whole JS context. On iOS that reads to the user as being logged
+ * out, since the core's WebView is what gets killed and reloaded.
+ *
+ * A cooldown bounds the cost to one worker per window, and the doubling keeps
+ * a long outage from spending any meaningful memory at all.
+ */
+const RETRY_BASE_MS = 30000
+const RETRY_MAX_MS = 300000
 
 /**
- * Initialize the NYM mixFetch client. Must be called before using mixFetch.
- * Safe to call multiple times - subsequent calls return the same promise.
+ * Builds the mixFetch setup routine over its own cooldown state.
+ *
+ * The returned function initializes the NYM mixFetch client, and must be
+ * called before using mixFetch. It is safe to call multiple times: subsequent
+ * calls return the same promise.
+ *
+ * A failed setup starts a cooldown: calls made before it expires fail
+ * immediately with the error that started it, instead of building another
+ * client.
+ *
+ * Tests build their own instance over a clock they control, which also gives
+ * them fresh state per case.
+ *
+ * The cooldown state is per instance, but the client, its worker and
+ * `window.__mixFetchGlobal` are process-global, so two live instances would
+ * each hold a cooldown of their own while spawning workers into the same
+ * process. The single instance exported below is the only supported
+ * arrangement.
  */
-export async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
-  if (mixFetchInitPromise == null) {
-    log('Initializing mixFetch...')
-    const pending = createMixFetch(mixFetchOptions)
-    // The timeout below can abandon this setup while it is still in flight.
-    // Deliberately do NOT tear it down on late completion: `createMixFetch`
-    // resolves to a healthy global singleton, and disconnecting it (a
-    // process-wide operation) would race a newer init that has taken over.
-    // A late completion just repopulates `__mixFetchGlobal`, which the next
-    // init reuses. Swallow a late rejection so it is not unhandled.
-    pending.catch(() => {})
-    let timer: ReturnType<typeof setTimeout> | undefined
-    const timeout = new Promise<never>((resolve, reject) => {
-      timer = setTimeout(() => {
-        reject(
-          new Error(`mixFetch setup timed out after ${SETUP_TIMEOUT_MS}ms`)
+export function makeMixFetchSetup(
+  now: () => number = () => Date.now()
+): (log: EdgeLog) => Promise<IMixFetchFn> {
+  let mixFetchInitPromise: Promise<IMixFetch> | null = null
+
+  /** When a new setup may be attempted, and the wait that produced it. */
+  let retryAfter: number = 0
+  let retryDelay: number = RETRY_BASE_MS
+
+  /** The failure to re-throw for callers that arrive during the cooldown. */
+  let lastError: unknown
+
+  return async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
+    if (mixFetchInitPromise == null) {
+      if (now() < retryAfter) {
+        // Re-throwing `lastError` itself would hand the caller a stack and a
+        // message from a setup that ended minutes ago, so a cooldown
+        // rejection reads in a crash report as a fresh 60-second timeout. It
+        // is also not necessarily an `Error`: the library rejects with a raw
+        // `MessageEvent` on a worker error, so `.message` can be undefined
+        // downstream.
+        //
+        // Nothing is logged here. This runs once per refused caller, and the
+        // breadcrumb that makes the quiet window visible is emitted once per
+        // window where the cooldown is armed.
+        const remainingMs = retryAfter - now()
+        const error: Error & { cause?: unknown } = new Error(
+          `mixFetch setup is cooling down for another ${Math.round(
+            remainingMs / 1000
+          )}s`
         )
-      }, SETUP_TIMEOUT_MS)
-    })
-    mixFetchInitPromise = Promise.race([pending, timeout])
-      .then(mixFetchModule => {
-        log('mixFetch initialized successfully')
-        return mixFetchModule
-      })
-      .catch(async error => {
-        // Clean up stale global state left by the failed init so the
-        // next createMixFetch call starts fresh instead of reusing a
-        // broken singleton.
-        try {
-          await disconnectMixFetch()
-        } catch {}
-        // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
-        delete (window as any).__mixFetchGlobal
-        mixFetchInitPromise = null
-        log.error('mixFetch initialization failed:', error)
+        error.cause = lastError
         throw error
+      }
+
+      log('Initializing mixFetch...')
+      const pending = createMixFetch(mixFetchOptions)
+      // The timeout below can abandon this setup while it is still in flight.
+      // `createMixFetch` publishes `window.__mixFetchGlobal` as soon as the
+      // worker exists and only then awaits the gateway handshake
+      // (`@nymproject/mix-fetch/index.js:446-449`), so the failure path
+      // usually deletes a global whose owner is already past that assignment:
+      // the abandoned worker stays alive and unreachable, and the next setup
+      // builds a fresh one. A delete that lands before the assignment lets the
+      // late completion re-publish instead, and the line right after it
+      // configures that client, so the next setup finds a working global
+      // rather than an unconfigured one. Either way the cost is one worker per
+      // cooldown window, against a ceiling of RETRY_MAX_MS. Swallow the late
+      // rejection so it is not unhandled.
+      pending.catch(() => {})
+      let timer: ReturnType<typeof setTimeout> | undefined
+      const timeout = new Promise<never>((resolve, reject) => {
+        timer = setTimeout(() => {
+          reject(
+            new Error(`mixFetch setup timed out after ${SETUP_TIMEOUT_MS}ms`)
+          )
+        }, SETUP_TIMEOUT_MS)
       })
-      .finally(() => {
-        clearTimeout(timer)
-      })
+      mixFetchInitPromise = Promise.race([pending, timeout])
+        .then(mixFetchModule => {
+          log('mixFetch initialized successfully')
+          return mixFetchModule
+        })
+        .catch(error => {
+          // Arm the cooldown before any cleanup. The timeout fires while the
+          // setup is still running, so the disconnect below is an RPC into a
+          // worker that is still busy and has no bounded completion; awaiting
+          // it would leave `mixFetchInitPromise` pending forever, and every
+          // later caller would await that instead of failing fast.
+          mixFetchInitPromise = null
+          lastError = error
+          retryAfter = now() + retryDelay
+          log.error(
+            `mixFetch initialization failed (no retry for ${Math.round(
+              retryDelay / 1000
+            )}s):`,
+            error
+          )
+          // One breadcrumb per cooldown window, not one per refused caller.
+          // The app caps Sentry at 25 breadcrumbs, so a poll loop retrying
+          // every few seconds would evict the whole history in about a
+          // minute, blinding the crash report this failure most needs to
+          // appear in. Emitted before the doubling below, so it names the
+          // window actually armed.
+          log.breadcrumb('mixFetch setup failed, cooling down', {
+            cooldownMs: retryDelay
+          })
+          retryDelay = Math.min(retryDelay * 2, RETRY_MAX_MS)
+
+          // Best-effort: clear the library's singleton so the next setup starts
+          // fresh instead of reusing a broken one. Not awaited, per above.
+          disconnectMixFetch().catch(() => {})
+          // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
+          delete (window as any).__mixFetchGlobal
+
+          throw error
+        })
+        .finally(() => {
+          clearTimeout(timer)
+        })
+    }
+    const mixFetchModule = await mixFetchInitPromise
+    return mixFetchModule.mixFetch
   }
-  const mixFetchModule = await mixFetchInitPromise
-  return mixFetchModule.mixFetch
 }
+
+export const initMixFetch = makeMixFetchSetup()
--- a/test/util/nym.test.ts
+++ b/test/util/nym.test.ts
@@ -1,0 +1,143 @@
+import { expect } from 'chai'
+import { afterEach, beforeEach, describe, it } from 'mocha'
+
+import { EdgeLog } from '../../src/types/types'
+import { expectRejection } from '../expect-rejection'
+
+/**
+ * `nym.ts` imports `@nymproject/mix-fetch`, whose entry point is browser-only
+ * WASM. These tests replace it in the module cache with a fake that counts how
+ * many clients get built, which is the number the cooldown exists to bound:
+ * every `createMixFetch` spawns a worker holding megabytes of WASM that the
+ * library never lets us terminate.
+ */
+let setupCount = 0
+let setupFails = true
+let disconnectHangs = false
+
+const mixFetchPath = require.resolve('@nymproject/mix-fetch')
+require.cache[mixFetchPath] = {
+  id: mixFetchPath,
+  filename: mixFetchPath,
+  loaded: true,
+  exports: {
+    async createMixFetch() {
+      ++setupCount
+      if (setupFails) throw new Error('gateway client error')
+      return { mixFetch: async () => new Response('') }
+    },
+    async disconnectMixFetch() {
+      // The real one is a Comlink RPC into the WASM worker, so it does not
+      // complete while that worker is still busy connecting.
+      if (disconnectHangs) await new Promise(() => {})
+    }
+  }
+} as unknown as NodeModule
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires
+const { makeMixFetchSetup } = require('../../src/util/nym')
+
+const breadcrumbs: string[] = []
+const fakeLog: EdgeLog = Object.assign(() => {}, {
+  breadcrumb: (message: string) => breadcrumbs.push(message),
+  crash: () => {},
+  warn: () => {},
+  error: () => {}
+})
+
+describe('nym mixFetch setup', function () {
+  let clock = 0
+  let initMixFetch: (log: EdgeLog) => Promise<unknown>
+
+  beforeEach(function () {
+    setupCount = 0
+    setupFails = true
+    disconnectHangs = false
+    breadcrumbs.length = 0
+    clock = 1_000_000
+    ;(global as any).window = {}
+    initMixFetch = makeMixFetchSetup(() => clock)
+  })
+
+  afterEach(function () {
+    delete (global as any).window
+  })
+
+  it('builds one client per cooldown window, not one per request', async function () {
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(1)
+
+    // Every caller during the cooldown fails without building a client. This
+    // is the leak the app hit: a poll loop drove a new WASM worker every few
+    // seconds until the host killed the JS context.
+    for (let i = 0; i < 10; ++i) {
+      clock += 2500
+      await expectRejection(initMixFetch(fakeLog))
+    }
+    expect(setupCount).equals(1)
+  })
+
+  it('rejects a cooling-down caller with its own error', async function () {
+    await expectRejection(initMixFetch(fakeLog))
+
+    // Not the original failure re-thrown: that reports a stale timeout in
+    // crash reports, and is not guaranteed to be an `Error` at all.
+    clock += 2500
+    const error = await initMixFetch(fakeLog).then(
+      () => undefined,
+      (error: Error & { cause?: unknown }) => error
+    )
+    expect(error?.message).contains('cooling down')
+    expect((error?.cause as Error)?.message).equals('gateway client error')
+  })
+
+  it('breadcrumbs once per cooldown window, not once per refused caller', async function () {
+    // The app caps Sentry at 25 breadcrumbs. One per refused caller would
+    // evict the app's whole crash-report history within a minute of polling.
+    await expectRejection(initMixFetch(fakeLog))
+    expect(breadcrumbs).deep.equals(['mixFetch setup failed, cooling down'])
+
+    for (let i = 0; i < 10; ++i) {
+      clock += 2500
+      await expectRejection(initMixFetch(fakeLog))
+    }
+    expect(breadcrumbs.length).equals(1)
+
+    // A fresh window is worth its own breadcrumb:
+    clock += 30000
+    await expectRejection(initMixFetch(fakeLog))
+    expect(breadcrumbs.length).equals(2)
+  })
+
+  it('retries once the cooldown expires, and backs off further', async function () {
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(1)
+
+    clock += 30000
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(2)
+
+    // The second failure doubles the wait, so the old 30s is not enough:
+    clock += 30000
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(2)
+
+    clock += 30000
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(3)
+  })
+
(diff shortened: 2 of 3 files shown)


src/util/nym.ts
```diff
@@ -32,15 +32,65 @@ export const mixFetchOptions: SetupMixFetchOps = {
...
  */
 export async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
   if (mixFetchInitPromise == null) {
+    if (now() < retryAfter) {
+      throw lastError ?? new Error('mixFetch setup is cooling down')
+    }
+
     log('Initializing mixFetch...')
     const pending = createMixFetch(mixFetchOptions)
     // The timeout below can abandon this setup while it is still in flight.
```
peachbits: The library doesn't behave the way this describes. `createMixFetch` assigns `window.__mixFetchGlobal` as soon as the **worker** is built, and only then awaits the gateway handshake:

```js
// node_modules/@nymproject/mix-fetch/index.js
if (!window.__mixFetchGlobal) {
  window.__mixFetchGlobal = await createMixFetch$1()   // worker exists here
  await window.__mixFetchGlobal.setupMixFetch(opts)    // this is what hangs
  ...
}
return window.__mixFetchGlobal
```

The `.catch` below deletes `__mixFetchGlobal` *after* that assignment, and `createMixFetch` never re-assigns it — so a late completion returns `undefined` rather than repopulating the global, and the next init builds a fresh worker instead of reusing the late one.

The cost of that is small, which is worth stating explicitly since the whole PR is about worker memory: the cooldown gates how often a new worker can be built at all, so the miss is at most one extra worker per window, and the window doubles to a 5 minute ceiling. That is the same order as the abandoned worker this path already accepts and leaves the per-request storm fixed — it makes the leak slow rather than eliminating it. The problem is just that the stated reason is wrong, and it is the kind of comment the next person will act on.

src/util/nym.ts
```diff
@@ -73,7 +126,15 @@ export async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
         // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
         delete (window as any).__mixFetchGlobal
         mixFetchInitPromise = null
-        log.error('mixFetch initialization failed:', error)
+        lastError = error
+        retryAfter = now() + retryDelay
```
peachbits: These lines are unreachable if the `await disconnectMixFetch()` a few lines up never settles, and the new timeout makes that newly possible. Before this PR the `.catch` only ran after `createMixFetch` itself rejected — the worker had already given up. Now it also fires at 60s while `setupMixFetch` is still in flight, and the disconnect is a Comlink RPC into that same busy WASM instance. `try { } catch {}` bounds a rejection, not a hang.

If it does hang, `mixFetchInitPromise` never settles and is never nulled, so every later `privacy: 'nym'` fetch awaits it forever and the cooldown never arms — the freeze this PR exists to prevent.

Suggest arming the cooldown first and firing the cleanup without awaiting it:

```ts
.catch(error => {
  mixFetchInitPromise = null
  lastError = error
  retryAfter = now() + retryDelay
  log.error(/* ... */)
  retryDelay = Math.min(retryDelay * 2, RETRY_MAX_MS)

  // Best-effort: clear the library's stale singleton so the next
  // createMixFetch starts fresh. Not awaited — a disconnect issued
  // against a still-connecting worker has no bounded completion.
  disconnectMixFetch().catch(() => {})
  // eslint-disable-next-line @typescript-eslint/no-dynamic-delete
  delete (window as any).__mixFetchGlobal

  throw error
})
```

That also drops the `async` from the handler. Related to line 96 — that block argues a process-wide disconnect is unsafe against an in-flight setup, which is exactly what this path does.

src/util/nym.ts
```diff
@@ -61,6 +111,9 @@ export async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
     mixFetchInitPromise = Promise.race([pending, timeout])
       .then(mixFetchModule => {
         log('mixFetch initialized successfully')
+        retryAfter = 0
+        retryDelay = RETRY_BASE_MS
+        lastError = undefined
```
peachbits: These three are unreachable. `mixFetchInitPromise` is only set back to `null` in the `.catch` below, so once a setup succeeds the guard at the top of `initMixFetch` is false forever and nothing ever reads `retryAfter` / `retryDelay` / `lastError` again.

Verified: deleting all three lines leaves the full suite at 166 passing, 0 failing.

Suggest deleting them along with the `clears the cooldown once a setup succeeds` test case. If clearing the backoff on success is meant to be real, it needs a path that invalidates a live client — a `mixFetch` call failing at runtime, say — and there isn't one today.

test/util/nym.test.ts
```diff
@@ -0,0 +1,105 @@
...
+    await initMixFetch(fakeLog)
+    expect(setupCount).equals(2)
+
+    // A later failure starts from the base wait again, not the backed-off one:
+    resetMixFetchForTests(() => clock)
+    setupFails = true
+    await expectRejection(initMixFetch(fakeLog))
+    clock += 30000
+    await expectRejection(initMixFetch(fakeLog))
+    expect(setupCount).equals(4)
```
peachbits: This case can't observe what it names, so it reads as coverage that isn't there. `resetMixFetchForTests` resets `retryDelay` itself, so the final assertion holds regardless of the success path; and in the first half `clock += 30000` lands exactly on `retryAfter`, so `now() < retryAfter` is already false and the retry would happen anyway. See the note on the `.then` block in `nym.ts` — the behavior is unreachable in the first place, so this should go with it.

src/util/nym.ts
```diff
@@ -32,15 +32,65 @@ export const mixFetchOptions: SetupMixFetchOps = {
...
+ */
+export function resetMixFetchForTests(
+  clock: () => number = () => Date.now()
+): void {
+  mixFetchInitPromise = null
+  retryAfter = 0
+  retryDelay = RETRY_BASE_MS
+  lastError = undefined
+  now = clock
+}
```
peachbits: Would rather not ship a `*ForTests` export and a mutable module clock in the library — it is the only one in `src/`, and the repo's existing idiom for an injectable clock is a parameter default (`src/util/crypto/hotp.ts` takes `now: number = Date.now() / 1000`).

Closing over the state gives the tests what they need without either. Production call sites and the function body stay exactly as they are:

```ts
export function makeMixFetchSetup(
  now: () => number = () => Date.now()
): (log: EdgeLog) => Promise<IMixFetchFn> {
  let mixFetchInitPromise: Promise<IMixFetch> | null = null
  let retryAfter = 0
  let retryDelay = RETRY_BASE_MS
  let lastError: unknown

  return async function initMixFetch(log) {
    // ...unchanged...
  }
}

export const initMixFetch = makeMixFetchSetup()
```

Both io files keep importing `initMixFetch` unchanged, and each test builds its own instance with `makeMixFetchSetup(() => clock)` — fresh state per test for free, no reset to call and nothing test-shaped in the shipped surface.

peachbits (commented):
The cooldown is the right shape and the doubling matches the changelog. Four notes: one comment that misdescribes the library, one hazard the new timeout opens up in the pre-existing cleanup block, and one set of resets in the success path that turn out to be unreachable (which is also why the third test passes either way).

src/util/nym.ts
j0ntz: Rewrote the comment to match the library. `createMixFetch` publishes `window.__mixFetchGlobal` as soon as the worker exists and only then awaits the handshake, so the failure path deletes a global that a late completion never re-assigns: that worker stays alive and unreachable, and the next setup builds a fresh one. The comment now says the cooldown is what bounds the cost, one worker per window against the `RETRY_MAX_MS` ceiling.

src/util/nym.ts
j0ntz: Took the restructure. The handler is no longer `async`: it nulls `mixFetchInitPromise`, records `lastError`, arms `retryAfter` and doubles `retryDelay` first, then fires `disconnectMixFetch().catch(() => {})` unawaited before deleting the global. Added a regression case that makes the fake disconnect never settle, which times out against the awaited form and passes with this one.

src/util/nym.ts
j0ntz: Deleted all three. `mixFetchInitPromise` is only nulled in the catch, so after a success the guard at the top is false forever and nothing reads them again.

test/util/nym.test.ts
j0ntz: Deleted, along with the resets it was meant to cover. The case that replaced it, `arms the cooldown even when the cleanup disconnect never settles`, was checked both ways: it times out against the previous awaited disconnect and passes against the fix.

src/util/nym.ts
j0ntz: Adopted as written: `makeMixFetchSetup(now = () => Date.now())` closes over the state and `export const initMixFetch = makeMixFetchSetup()` keeps both io call sites unchanged. `resetMixFetchForTests` and the module-level clock are gone, and each test now builds its own instance with `makeMixFetchSetup(() => clock)`.

j0ntz:
All four taken, no rejections. The comment at the top of the setup now describes what the library actually does, the failure handler arms the cooldown before firing an unawaited disconnect, and the unreachable success-path resets are gone with the test case that appeared to cover them.

The hang you flagged is now covered: a new case makes the fake `disconnectMixFetch` never settle, which times out against the awaited form and passes against the fix. Suite is 166 passing, same as before, and `tsc --noEmit` is clean.

src/util/nym.ts
peachbits: **Retracting the premise of my earlier note here.** I claimed the 60s timeout was new in this PR and therefore newly routed an in-flight setup into the cleanup block. That's wrong — `SETUP_TIMEOUT_MS`, the `Promise.race`, and the `await disconnectMixFetch()` are all at `HEAD~1:src/util/nym.ts:33,53-61,71`, shipped in 2.47.1 (*"Bound the NYM mixFetch setup at 60 seconds"*). This PR neither introduces nor worsens it.

The hazard itself is real and now better supported than I had it. Comlink's request path is `index.js:342-351`:

```js
function requestResponseMessage(ep, pendingListeners, msg, transfers) {
  return new Promise((resolve) => {   // no reject, no timeout
```

So a disconnect issued against a worker that has been killed — the memory-pressure case this PR is about — never settles. I confirmed the consequence empirically by stubbing `disconnectMixFetch` to hang against the real module: the first call hangs, and so does a later call with the clock advanced 10,000,000 ms past any cooldown, because `mixFetchInitPromise = null` sits downstream of that `await` and never runs.

So: pre-existing, low probability, permanent for the life of the JS context, and correlated with exactly the failure this PR targets. Your call whether to fix it here, but the catch block is already open — arming the cooldown first and firing `disconnectMixFetch().catch(() => {})` without awaiting is a two-line change.

src/util/nym.ts
peachbits: Same scoping error — this comment is pre-existing (`HEAD~1:src/util/nym.ts:46-51`), not part of this PR. The reading of the library still holds, and one correction makes it sharper than what I wrote.

I said no path repopulates `__mixFetchGlobal`. That's too absolute. If the timeout fires while the zombie is still inside `createMixFetch$1()` — worker boot plus 12.8 MB of WASM, which is plausible under the memory pressure this PR is about — its `window.__mixFetchGlobal = await createMixFetch$1()` lands *after* our `delete`, and the global is repopulated with a worker that never had `setupMixFetch` called on it. The next init then hits `if (!window.__mixFetchGlobal)` at `index.js:446`, finds it truthy, and returns it **without running setup**, so `initMixFetch` logs success and hands callers an unconfigured client.

Narrow, and again not something this PR introduces — but it means the comment's conclusion is wrong in both directions, so it is worth rewriting rather than trusting.

src/util/nym.ts
```diff
@@ -32,15 +32,65 @@ export const mixFetchOptions: SetupMixFetchOps = {
...
+ *
+ * A failed setup starts a cooldown: calls made before it expires fail
+ * immediately with the error that started it, instead of building another
+ * client.
  */
 export async function initMixFetch(log: EdgeLog): Promise<IMixFetchFn> {
   if (mixFetchInitPromise == null) {
+    if (now() < retryAfter) {
+      throw lastError ?? new Error('mixFetch setup is cooling down')
+    }
```
peachbits: Two things about this throw, on code the PR does add:

- **It re-throws the identical instance.** A caller rejected in 0 ms gets the original object, stack and all — commonly `mixFetch setup timed out after 60000ms`. Crash reports and logs will misattribute a cooldown rejection as a fresh 60-second timeout. And `lastError` is not necessarily an `Error`: `index.js:387` rejects with a raw `MessageEvent` (`reject(msg)`), so `error.message` can be `undefined` downstream.
- **The cooldown is silent.** The failure logs once at line 131; every rejection for the next 30s-5min emits nothing. For a change whose whole purpose is making a user-visible "logged out" symptom diagnosable, a `log.breadcrumb` here would be worth the noise.

src/util/nym.ts
peachbits: Two corrections to my earlier note, neither of which changes the suggestion:

- Drop my `hotp.ts` citation. It injects a *timestamp* (`now: number = Date.now() / 1000`), not a clock, and it holds no cross-call state — this module reads the clock at two points separated by event-loop turns, so the analogy doesn't transfer. `makeMixFetchSetup` stands on its own merits.
- I said the export doesn't reach the public API. More precisely: `build.lib` transpiles all of `src/`, and `files[]` includes `/lib/`, so `resetMixFetchForTests` does ship in the tarball at `lib/util/nym.js`. It's unreachable through the `exports` map, not absent.

One caveat worth stating on the factory itself: closure state makes the cooldown look per-instance while `window.__mixFetchGlobal` and its workers stay process-global, so two instances would each think they own a cooldown while spawning workers into the same process. Fine for one production instance; worth a comment if you take the refactor.

peachbits (commented):
Correction to my earlier review. I diffed against a stale local `master` (2.46.1), which folded 2.47.x/2.48.0 into what I read as this PR's diff. Three notes therefore described pre-existing code as new — one of them on a factually wrong premise. Retracting that one and re-scoping the rest below. The notes on the `.then` resets (114-116), the third test case, and the `*ForTests` export were about code this PR actually adds and stand as written.

src/util/nym.ts
j0ntz: Took it: the cooldown is now armed before the library is touched, and the disconnect is fired without awaiting (`disconnectMixFetch().catch(() => {})`). Confirmed your reading of `requestResponseMessage` in the installed copy, it resolves only on a reply with no reject path and no timeout, so awaiting it against a killed worker stranded every line below it, cooldown included. `test/util/nym.test.ts` now has a case with the fake disconnect hanging forever: it fails on the old ordering by timing out and passes on the new one, with the retry after the cooldown still firing.

src/util/nym.ts
j0ntz: Rewrote the comment, though the replacement says something different from both versions. Tracing the installed `index.js:446-449`, the assignment and the setup are inside one `createMixFetch` call (`window.__mixFetchGlobal = await createMixFetch$1()` then `await window.__mixFetchGlobal.setupMixFetch(opts)`), and line 449 re-reads the global rather than a local, so no landing position of our `delete` leaves a configured-looking global that never had setup run:

```mermaid
flowchart TD
  Q{"Where does the failure path's delete land?"}
  Q -->|before 448| A["448 re-assigns, 449 configures it: present and configured"]
  Q -->|between 448 and 449| B["449 re-reads a deleted global: TypeError, pending rejects, global stays gone"]
  Q -->|after 449| C["a configured client is discarded: worker leaked, next init rebuilds"]
```

The truthy-but-unconfigured global does exist, but it comes from `setupMixFetch` itself rejecting after 448 already assigned, which is the case the `delete` in the catch is there for. The new comment states that and keeps the leaked worker as the real cost of an abandoned setup.

src/util/nym.ts
j0ntz: Both taken. The cooldown now rejects with its own error naming the remaining wait, carrying the original as `cause` rather than re-throwing the instance, which also removes the assumption that `lastError` is an `Error` at all. And it emits a `log.breadcrumb` with the remaining milliseconds on every refused call, so the quiet window is visible in a crash report instead of ending at the single failure line. New test asserts the message and the preserved cause.

src/util/nym.ts
j0ntz: Not taking the factory, on the strength of your own caveat: the client, its worker and `window.__mixFetchGlobal` are process-global, so closure state would advertise a per-instance cooldown over shared workers, which is a worse lie than the module-level state telling the truth. Added a comment at the state block saying exactly that, so the next reader does not re-derive the refactor. Noted on `resetMixFetchForTests` shipping in `lib/` but unreachable through the `exports` map, and on the `hotp.ts` citation.

j0ntz:
Re-scoping noted, and the three re-scoped notes are answered in their threads. Two produced code: the catch block now arms the cooldown before firing an un-awaited `disconnectMixFetch`, and the cooldown rejects with its own error plus a breadcrumb instead of re-throwing the stored instance. The `__mixFetchGlobal` comment was rewritten, but against a third reading, traced through `index.js:446-449` in the installed copy, which does not match either of ours. The `makeMixFetchSetup` factory is declined, with the reason recorded in the code.

src/util/nym.ts
j0ntz: Correcting my previous reply: it described a stale checkout. `makeMixFetchSetup` is already on the branch and has been since the last round, so the factory was taken, not declined. What was actually missing is your caveat, which is now a paragraph on the factory's doc comment: the cooldown state is per instance while the client, its worker and `window.__mixFetchGlobal` are process-global, so the single exported instance is the only supported arrangement.

src/util/nym.ts
```diff
@@ -32,54 +32,139 @@ export const mixFetchOptions: SetupMixFetchOps = {
...
+        // Re-throwing `lastError` itself would hand the caller a stack and a
+        // message from a setup that ended minutes ago, so a cooldown
+        // rejection reads in a crash report as a fresh 60-second timeout. It
+        // is also not necessarily an `Error`: the library rejects with a raw
+        // `MessageEvent` on a worker error, so `.message` can be undefined
+        // downstream. The breadcrumb is what makes the quiet window visible,
+        // since the failure itself logs only once.
+        const remainingMs = retryAfter - now()
+        log.breadcrumb('mixFetch setup refused, still cooling down', {
+          remainingMs
```
peachbits: My fault — I asked for a breadcrumb without saying where. Once per refused caller is the wrong cardinality, and the numbers are worse than I'd have guessed.

`EdgeCrashReporter.logBreadcrumb` in the GUI (`src/components/services/EdgeCoreManager.tsx:92`) is a straight pass-through to `addBreadcrumb` from `@sentry/react-native`, and that same object is handed to `MakeEdgeContext`, so this call lands in Sentry from inside the core WebView. The GUI sets **`maxBreadcrumbs: 25`** (`src/app.ts:45`) — the SDK default is 100, so it was tuned down deliberately, alongside `breadcrumbsIntegration({ console: false })`. It's a hard ring buffer, trimmed oldest-first in `@sentry/core/build/cjs/breadcrumbs.js`.

At the cadence this PR describes — every few seconds, 2.5s in its own test — 25 refusals evict the app's entire breadcrumb history in about **a minute**, and a 5-minute ceiling window emits ~120, cycling the buffer roughly five times. A crash arriving during a NYM outage would carry 25 identical `still cooling down` entries and none of the 19 one-per-user-action breadcrumbs (`loginWithPassword`, `createWallet`, `makeKeysKit`…). That crash report is the artifact for the iOS JS-context kill this PR fixes, so the breadcrumb blinds the report it exists to inform.

Moving it to the arming site keeps the window visible at one breadcrumb per window. Drop the call here, and in the `.catch`, between `log.error` and the doubling so it names the window actually armed:

```ts
// One breadcrumb per cooldown window, not one per refused caller.
// The app caps Sentry at 25 breadcrumbs, so a poll loop retrying
// every few seconds would evict the whole history in about a
// minute, blinding the crash report this failure most needs to
// appear in. Emitted before the doubling below, so it names the
// window actually armed.
log.breadcrumb('mixFetch setup failed, cooling down', {
  cooldownMs: retryDelay
})
retryDelay = Math.min(retryDelay * 2, RETRY_MAX_MS)
```

With a guard alongside it, since the failure mode is silent:

```ts
it('breadcrumbs once per cooldown window, not once per refused caller', async function () {
  await expectRejection(initMixFetch(fakeLog))
  expect(breadcrumbs).deep.equals(['mixFetch setup failed, cooling down'])

  for (let i = 0; i < 10; ++i) {
    clock += 2500
    await expectRejection(initMixFetch(fakeLog))
  }
  expect(breadcrumbs.length).equals(1)

  // A fresh window is worth its own breadcrumb:
  clock += 30000
  await expectRejection(initMixFetch(fakeLog))
  expect(breadcrumbs.length).equals(2)
})
```

(needs `fakeLog.breadcrumb` to push into a `breadcrumbs: string[]` cleared in `beforeEach`.)

I have this as a fixup locally — 168 passing, `tsc` and `eslint` clean, and putting the breadcrumb back on the per-caller path fails the new test with `expected 11 to equal 1`. Happy to hand it over, or it's a two-line move.

j0ntz: Taken as you pushed it: `d71fac64` is on the branch, so the breadcrumb is gone from the refusal path and now fires in the `.catch` between `log.error` and the doubling, carrying `cooldownMs: retryDelay` so it names the window actually armed rather than the next one. Confirmed the premise independently in the GUI: `src/app.ts:45` sets `maxBreadcrumbs: 25`, and `EdgeCoreManager.tsx:94` passes `logBreadcrumb` straight through to Sentry. The cardinality guard holds on that head, where the suite reports 168 passing, `tsc` is clean, and `verify-repo.sh` passed.

j0ntz:
## 📸🪓 Test evidence: cooldown driven in the running app

> 🪓 **Hack-forced evidence:** the gateway failure is forced by a temporary uncommitted edit that rejects the setup after 2s (the real client is still built, so each attempt still spawns its worker); the second column additionally sets RETRY_BASE_MS=0 to reproduce the pre-fix cadence. Both edits reverted, worktree clean. Temporary uncommitted edit, reverted before commit; the marked frames prove the rendering, not the trigger.

**🪓 HACK-FORCED: nym mixnet enabled**

🪓 HACK-FORCED: nym mixnet enabled

**🪓 HACK-FORCED: nym cooldown ab**

🪓 HACK-FORCED: nym cooldown ab

_Captured by the agent's in-app test run (build-and-test)._

peachbits (approved):

PR #739, opened 2026-08-29
Add EdgeContextOptions.apiSigner for native HMAC delegation

paullinator:
### CHANGELOG

- [x] Yes
- [ ] No

### Dependencies

- https://github.com/EdgeApp/edge-react-gui/pull/6183
- https://github.com/EdgeApp/edge-info-server/pull/163

### Description

Adds `EdgeContextOptions.apiSigner` so native code can sign Edge API HMAC requests instead of holding `apiSecret` in the JS bundle. GUI uses this for the native HMAC signer; JS `apiKey`/`apiSecret` remain a fallback.

### DIFF
--- a/docs/api-signer.md
+++ b/docs/api-signer.md
@@ -1,0 +1,78 @@
+# API request signing (`apiSigner`)
+
+`makeEdgeContext` / `MakeEdgeContext` can delegate HMAC for **login-server**
+requests so the HMAC secret never enters the JS bundle.
+
+This is a login-server contract only (existing HMAC in
+`edge-login-server/src/middleware/with-api-key.ts`). The GUI’s `GET /v1/getKeys`
+call is signed in the app, not by this library. See
+[edge-react-gui `docs/HMAC_SIGNING.md`](https://github.com/EdgeApp/edge-react-gui/blob/develop/docs/HMAC_SIGNING.md).
+
+## `EdgeContextOptions.apiSigner`
+
+```ts
+interface EdgeApiSignature {
+  apiKey: string      // public id for the Authorization header
+  signature: string   // base64 HMAC-SHA256 of the message
+}
+
+interface EdgeApiSigner {
+  signMessage: (message: string) => Promise<EdgeApiSignature>
+}
+
+interface EdgeContextOptions {
+  apiKey?: string
+  apiSecret?: Uint8Array
+  apiSigner?: EdgeApiSigner  // takes precedence over apiKey / apiSecret
+  appId: string
+  // ...
+}
+```
+
+On React Native, pass the same `apiSigner` prop to `MakeEdgeContext`. The
+bridge `bridgifyObject`s it and the WebView worker forwards it into
+`makeContext`. Implementors must return a usable `apiKey` (non-empty, no
+whitespace) and a non-empty signature.
+
+## Canonical string
+
+`loginFetchInner` builds the UTF-8 message the signer (or `apiSecret`) HMACs:
+
+```
+{METHOD}\n/api{path}\n{BODY}
+```
+
+- `METHOD` is the HTTP method (`POST`, `GET`, …).
+- Path is `/api` plus the login route, including any query string
+  (`/api/v2/login`, `/api/v2/login/create`, …).
+- `BODY` is `JSON.stringify(wasLoginRequestBody(body))`, or empty for GET /
+  omitted bodies.
+
+The Authorization header is:
+
+```
+HMAC {apiKey} {signature}
+```
+
+When `apiSigner` is set, its `apiKey` and `signature` are used even if
+`apiKey` / `apiSecret` were also passed. When `apiSigner` is absent and
+`apiSecret` is present, the core HMACs with that secret. When neither is
+present, the core sends the legacy `Token {apiKey}` header.
+
+There is **no** timestamp line and **no** `X-Timestamp` header. That extra line
+is info-server `getKeys` only; do not feed a four-line getKeys string into this
+signer for login, or a three-line login string into getKeys.
+
+## Attestation (separate from HMAC)
+
+`EdgeContext.setAttestationToken(jwt | undefined)` copies a short-lived
+info-server attestation JWT onto subsequent login-server requests as
+`x-attestation-token`. It does not participate in HMAC. A missing or invalid
+token does not change how this library signs; the login server may treat it as
+unattested and continue. getKeys (GUI → info-server) 401s on a bad token
+instead.
+
+## Tests
+
+`test/core/login/api-signer.test.ts` checks that `apiSigner` wins over
+`apiSecret` and that the signed message starts with `POST\n/api/`.
--- a/src/core/login/login-fetch.ts
+++ b/src/core/login/login-fetch.ts
@@ -1,6 +1,7 @@
-import { asMaybe } from 'cleaners'
+import { asMaybe, asObject, asString, Cleaner } from 'cleaners'
 import { base64 } from 'rfc4648'
 
+import { ApiSignerError, asMaybeApiSignerError } from '../../types/error'
 import {
   asChallengeErrorPayload,
   asLoginResponseBody,
@@ -9,6 +10,8 @@ import {
 import { LoginRequestBody } from '../../types/server-types'
 import {
   ChallengeError,
+  EdgeApiSignature,
+  EdgeApiSigner,
   EdgeFetchOptions,
   EdgeFetchResponse,
   NetworkError,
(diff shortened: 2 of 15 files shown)


j0ntz (commented):
Reviewed alongside edge-react-gui#6183 and edge-info-server#163. The `apiSigner` contract itself is clean: precedence over `apiKey`/`apiSecret` is explicit, `ApiSignerError` breaks out of the per-server retry loop at `login-fetch.ts:145` instead of being treated as a network failure, and `bridgifyObject` is applied on the RN path with the fake-world case handled GUI-side.

**`checkPin2` erases `ApiSignerError`** (outside the diff, so it lives here). `src/core/login/pin2.ts:186` ends with `.then(good => true, bad => false)`, so a signer timeout or a malformed native signature during PIN login reaches `account.checkPin` (`account-api.ts:372`) as "wrong PIN". The `bad => false` predates this PR, but before it every rejection there was either a network error or a bad PIN. `ApiSignerError` exists so signing failures stay distinguishable, and this is the one path that discards that. Rethrowing it before the collapse restores the distinction:

```ts
.then(good => true, (bad: unknown) => {
  if (asMaybeApiSignerError(bad) != null) throw bad
  return false
})
```

Also: `8f82eb9a` has a 60-character subject (cap is 50).

src/core/login/login-fetch.ts
```diff
@@ -22,6 +25,48 @@ import { utf8 } from '../../util/encoding'
 import { timeout } from '../../util/promise'
 import { ApiInput } from '../root-pixie'
 
+const asEdgeApiSignature: Cleaner<EdgeApiSignature> = asObject({
+  apiKey: asString,
+  signature: asString
+})
+
+function isUsableSignerKey(apiKey: string, signature: string): boolean {
+  return apiKey !== '' && !/\s/.test(apiKey) && signature !== ''
```
j0ntz, line 34: `apiKey` is checked for whitespace, `signature` only for emptiness, and both are interpolated into the same header value on the next lines.

The old in-JS path could not produce a bad signature because it was always `base64.stringify(hmacSha256(...))`. An `apiSigner` is by definition external code, so this is the first place an arbitrary string reaches `Authorization`. A signature carrying a space, tab, or CR/LF passes this check and yields a header the login server cannot parse as `HMAC key sig`, which reads as a signing mismatch rather than a malformed header.

```ts
const isUsable = (value: string): boolean => value !== '' && !/\s/.test(value)
function isUsableSignerKey(apiKey: string, signature: string): boolean {
  return isUsable(apiKey) && isUsable(signature)
}
```

edge-react-gui#6183 has the same asymmetry at `src/util/edgeApiSigner.ts:73`.

src/react-native.tsx
```diff
@@ -185,13 +187,14 @@ export async function fetchLoginMessages(
...
 
-  // API key:
-  let authorization = `Token ${apiKey}`
-  if (apiSecret != null) {
-    const requestText = `POST\n/api/v2/messages\n${bodyText}`
-    const hash = hmacSha256(utf8.parse(requestText), apiSecret)
-    authorization = `HMAC ${apiKey} ${base64.stringify(hash)}`
-  }
+  // Authorization:
+  const requestText = `POST\n/api/v2/messages\n${bodyText}`
```
j0ntz, line 191: The canonical string is now built in two places. `loginFetchInner` uses `${method}\n/api${path}\n${bodyText ?? ''}` (`login-fetch.ts:174`); this line hardcodes the same shape for `/v2/messages`.

`docs/api-signer.md` documents that format as one contract owned by `loginFetchInner`, so a future change made there is unlikely to also be made here. The failure mode is narrow but confusing: login-alert fetches start 401ing while ordinary login keeps working, which points at the server rather than a stale local template. Since this PR is already centralizing header construction into `makeLoginAuthorization`, a sibling `buildLoginRequestText(method, path, bodyText)` exported from the same file and called from both sites closes it.

src/core/plugins/plugins-selectors.ts
```diff
@@ -60,29 +60,16 @@ export function getCurrencyTools(
...
-      }
-    }
-    if (missingPlugins.length > 0) {
-      throw new Error(
-        'The following plugins are missing or failed to load: ' +
-          missingPlugins.join(', ')
-      )
-    }
-
     return true
```
j0ntz, line 73: Agreed with the change (one unusable plugin should not block every login), but the diagnostic went away entirely rather than moving down a severity level. A plugin that fails to load is now absent from `currencyConfig`/`swapConfig` with no error, warning, or log anywhere in the login path.

edge-react-gui#6183 compensates with `reportMissingPlugins` in `LoginActions.tsx`, so the Edge app is covered. Other consumers of this library are not, and this is exactly the case where a typo'd `pluginId` used to be impossible to miss. A `props.log.warn` listing the ids that were requested but never appeared keeps the signal at a severity that cannot take down a login:

```ts
const missing = Object.keys(init).filter(
  id => init[id] !== false && init[id] != null && currency[id] == null && swap[id] == null
)
if (missing.length > 0) props.log.warn(`Plugins failed to load: ${missing.join(', ')}`)
```

docs/api-signer.md
```diff
@@ -0,0 +1,78 @@
+# API request signing (`apiSigner`)
+
+`makeEdgeContext` / `MakeEdgeContext` can delegate HMAC for **login-server**
+requests so the HMAC secret never enters the JS bundle.
+
+This is a login-server contract only (existing HMAC in
+`edge-login-server/src/middleware/with-api-key.ts`). The GUI’s `GET /v1/getKeys`
```
j0ntz, line 7: `GET /v1/getKeys` does not exist. The route the GUI signs is `GET /v1/infoRollup/:appId` (edge-info-server `src/routes/v1/infoRollupRoutes.ts:25`, mounted at `/v1` in `src/indexInfo.ts:44`). `grep -rn getKeys` finds zero route hits in edge-info-server's `src/routes/` and `src/public/`, and zero in the GUI.

The stale name also appears at lines 63, 64, and 72 of this file ("a four-line getKeys string", "getKeys (GUI to info-server) 401s on a bad token"). Since this doc's purpose is keeping the two signing formats apart, naming the wrong endpoint for one of them is the error worth catching here.

src/core/login/login-fetch.ts
j0ntz: Addressed in 05c665ad: the signature is now rejected for whitespace as well as emptiness, same as the key.

src/react-native.tsx
j0ntz: Addressed in 2b9b4528: both callers build the string with `makeLoginRequestText`, and docs/api-signer.md points new signed routes at it.

src/core/plugins/plugins-selectors.ts
j0ntz: Addressed in bffed89d: a plugin still switched on but missing from both maps now logs a warning. One follow-up on the new comment there is in this review.

docs/api-signer.md
j0ntz: Addressed in ff316c58: the doc now names `GET /v1/infoRollup/:appId`.

src/core/plugins/plugins-selectors.ts
```diff
@@ -60,29 +60,34 @@ export function getCurrencyTools(
...
   })
+
+  // A plugin whose factory threw is logged where it failed, and the reducer
```
j0ntz: The comment says a plugin whose factory threw has its `init` entry switched off by the reducer. It does not. In `plugins-actions.ts:96` the catch only logs, so `out[pluginId]` is never set and the reducer's `plugin == null -> init: false` branch (`plugins-reducer.ts:44`) never sees that id. A plugin that threw therefore stays `init` on, lands in `missing`, and is logged a second time as "Requested plugins are not installed", which points the reader at a misspelled id rather than the factory error. Either set `out[pluginId] = undefined` in the catch (so the reducer records it as failed) or fix the comment and message.

src/io/react-native/react-native-worker.ts
```diff
@@ -261,10 +261,15 @@ export function normalizePath(path: string): string {
...
-  async makeEdgeContext(nativeIo, logBackend, pluginUris, opts) {
+  async makeEdgeContext(nativeIo, logBackend, pluginUris, opts, apiSigner) {
     loadPlugins(pluginUris)
     const io = await makeIo(logBackend)
-    return await makeContext({ io, nativeIo }, logBackend, opts)
+    // The signer travels as its own bridged argument, but honor one that
+    // arrived inside `opts` rather than silently dropping it.
+    return await makeContext({ io, nativeIo }, logBackend, {
+      ...opts,
+      apiSigner: apiSigner ?? opts.apiSigner
```
j0ntz: `apiSigner ?? opts.apiSigner` is unreachable on the only real caller: `src/react-native.tsx:57` destructures `apiSigner` out of the props and sends it as the separate bridged argument, so `opts` never carries it. A function inside the plain `opts` object would not survive yaob anyway, so the comment ("honor one that arrived inside opts rather than silently dropping it") describes a case that cannot work. Suggest dropping the fallback and the comment.

docs/api-signer.md
```diff
@@ -0,0 +1,84 @@
...
+  apiSigner?: EdgeApiSigner  // takes precedence over apiKey / apiSecret
+  appId: string
+  // ...
+}
+```
+
+On React Native, pass the same `apiSigner` prop to `MakeEdgeContext`. The
+bridge `bridgifyObject`s it and the WebView worker forwards it into
+`makeContext`. Implementors must return a usable `apiKey` (non-empty, no
+whitespace) and a non-empty signature.
```
j0ntz: "a non-empty signature" is looser than the code: `isUsableSignerKey` in `login-fetch.ts` also rejects whitespace in the signature. Suggest "a non-empty signature with no whitespace" to match the apiKey sentence.

j0ntz (approved):

